Issue #10, opened 2021-09-21
Wrong filename / missing images produces cryptic openCV error message


hansent:
I tried uploading an image using the pip package and accidentally misspelled the filename.  When an image doesn't exists I get the following error when e.g. trying to call: `project.upload("/Users/hansent/Desktop/original.jpeg")`

Error:
```
Traceback (most recent call last):
  File "/Users/hansent/Desktop/upload.py", line 16, in <module>
    project.upload("/Users/hansent/Desktop/original.jpg")
  File "/usr/local/lib/python3.9/site-packages/roboflow/core/project.py", line 143, in upload
    response = self.__image_upload(image_path, hosted_image=hosted_image, split=split)
  File "/usr/local/lib/python3.9/site-packages/roboflow/core/project.py", line 89, in __image_upload
    image = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
cv2.error: OpenCV(4.5.3) /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-req-build-vy_omupv/opencv/modules/imgproc/src/color.cpp:182: error: (-215:Assertion failed) !_src.empty() in function 'cvtColor'

```

It would be nice if the library checked for the existence of the file and gave a nicer error message instead.

Palashio:
fixed this in PR, check it out!

PR #12, opened 2021-09-21
fixes bug #10

Palashio:


Jacobsolawetz:
looks good to me, glad we don't leave you with a cryptic OpenCV error anymore

PR #20, opened 2021-11-23
Add more warnings when upload fails (which happens 20% of the time). …

brianegge:
…Also add a return code to the upload_single so clients can do something on failure


yeldarby:
Thanks Brian! We were in a weird state with a partial refactor on a different branch. We'll work on getting these changes worked into `main` soon!

brianegge:
Having a return code for single_upload was essential because it failed so often. I downloaded my private workspace, blurred the faces using mtcnn, and then re-uploaded with the following script
```
for filename in os.listdir(directory):
    if filename.endswith(".jpg"):
        f = os.path.join(directory, filename)
        a = f.replace(".jpg", ".xml")
        print(f"Uploading {f}")
        success = project.single_upload(image_path=f, annotation_path=a, split=split)
        if not success:
            print(f"Retrying {f}")
            success = project.single_upload(image_path=f, annotation_path=a, split=split)
```

PR #23, opened 2022-03-11
update image uploading to multipart forms

patrick-morgan:
Replaced the base64 encoding of images with a multipart/form-data POST request

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -10,10 +10,11 @@ import requests
 from PIL import Image
 from roboflow.config import *
 from roboflow.core.version import Version
+from requests_toolbelt.multipart.encoder import MultipartEncoder
 
 #version class that should return
 class Project():
-    def __init__(self, api_key, a_project, model_format):
+    def __init__(self, api_key, a_project, model_format=None):
         if api_key == "coco-128-sample":
             self.__api_key = api_key
             self.model_format = model_format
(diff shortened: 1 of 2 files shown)


roboflow/core/project.py
```diff
@@ -101,30 +102,13 @@ def __image_upload(self, image_path, hosted_image=False, split="train"):
...
             self.image_upload_url = "".join([
                 "https://api.roboflow.com/dataset/", project_name, "/upload",
-                "?api_key=", self.__api_key,
-                "&name=" + image_name,
-                "&split=" + split
+                "?api_key=", self.__api_key
             ])
 
-            # Convert to PIL Image
-            img = cv2.imread(image_path)
```
hansent, line 113: wondering if we may be changing (undocumented?) behavior here:

Looks like before we were reading the image with PIL and then "saving" to JPEG before base64 encoding; whereas now we just upload the image directly as is on the local filesystem.

I don't think our API handles conversion, so may be safer to leave the PIL loading and JPEG conversion in?  I could imagine e.g. a customer currently using this to upload a folder of png, tiff or any other PIL supported image format and then it just magically working and then breaking their workflow by changing this to upload the file without converting to JPEG.

hansent: @yeldarby are the images we save to GCS always jpeg?

yeldarby: Not sure; what do we do in the frontend? I think we upload them as-is & convert them as part of `roboflow-transform`.

We should verify that they are images for sure though; don't want someone uploading a malicious exe & serving it from our bucket or something.

patrick-morgan: Good catch, so we want to leave the PIL loading and JPEG conversion and then send that JPEG in the multipart form?

hansent: > so we want to leave the PIL loading and JPEG conversion and then send that JPEG in the multipart form
yes, I think that would be ideal

hansent (approved):
lgtm.  Need to figure out how to publish this

roboflow/core/project.py
```diff
@@ -101,31 +102,23 @@ def __image_upload(self, image_path, hosted_image=False, split="train"):
             image_name = os.path.basename(image_path)
 
             # Construct URL for local image upload
-
             self.image_upload_url = "".join([
                 "https://api.roboflow.com/dataset/", project_name, "/upload",
-                "?api_key=", self.__api_key,
-                "&name=" + image_name,
-                "&split=" + split
```
Jacobsolawetz, line 109: @patrick-morgan @hansent did we ablate split and name because things changed in the upload api with the new label flow? otherwise i'll add them back in before deploy

Jacobsolawetz (approved):
Looks good to me - awaiting one minor comment before merge

Jacobsolawetz:
Dogfooded by uploading these shrimp and this mushroom 
image

roboflow/core/project.py
patrick-morgan: Yeah, those parameters are no longer used in the URL for multipart form uploading. That data (split and name) is now sent in the request on line 120. Good that you pointed that out though as I will update the documentation to make that more clear

PR #24, opened 2022-03-23
respect API_URL config parameter for all API calls

hansent:
The image and annotation upload functions were using. hardcoded endpoint instead of the `API_URL` config parameter; which makes it hard to use for testing in development environments.


Jacobsolawetz (approved):
Nice

PR #25, opened 2022-03-29
Active learning in roboflow pip

Jacobsolawetz:
Jay and Charles have made progress on adding active learning utility scripts to the pip package. I believe these are an extension of techniques already used in field, via the active learning notebook

```
Active learning tends to leverage rules that fall into the following buckets:

Target class: "only upload images which contain a specific class type"
Confidence interval: "only upload images which fall into a specific confidence range"
Intra-class variance bounds: "only upload images which contain more/less instances of a specific detection class"
Object size and occlusion proxy bounds: "only upload images with a bounding box of certain sizes"
Similarity comparison, only grab images dissimilar[link text](https://github.com/roboflow-ai/roboflow-python/pull/25) to previous images to avoid unneccsary duplicates
```

Genesis notebook --> https://colab.research.google.com/drive/1aR3F34zHUN6j5cAIdP1OrDXXOdiLB805#scrollTo=INzAGQMw3YY0


roboflow/core/workspace.py
```diff
@@ -61,6 +62,64 @@ def project(self, project_name):
...
+        '''
+
+        inference_model = self.project(inference_endpoint[0]).version(inference_endpoint[1]).model
+        upload_project = self.project(upload_destination)
+
+        print("inference reference point: ", inference_model)
+        print("upload destination: ", upload_project)
+
+        # TODO: work with raw_data
+        # TODO: extention and globbing properties added to config
```
Jacobsolawetz, line 81: +1

Jacobsolawetz: We should make the globbing a little more flexible

roboflow/core/workspace.py
```diff
@@ -61,6 +62,64 @@ def project(self, project_name):
...
+                
+                    # filter out non-target_class uploads if enabled
+                    if(len(conditionals["target_classes"]) > 0 and
+                        prediction['class'] not in conditionals["target_classes"]):
+                        print(' [X] prediction failed target_classes')
+                        continue
+
+                    # upload on success!
+                    print(' >> image uploaded!')
+                    upload_project.upload(image, num_retry_uploads=3)
```
Jacobsolawetz, line 118: Does this attempt to upload multiple instances of the image if multiple prediction boxes pass the logic? Seems like we should just track a flag in the for loop and upload once

Jacobsolawetz:
test comment

roboflow/core/workspace.py
Jacobsolawetz: Nope - you had the break at the end, rad

roboflow/core/workspace.py
Jacobsolawetz: ✅

roboflow/core/workspace.py
```diff
@@ -61,6 +62,76 @@ def project(self, project_name):
 
         return Project(self.__api_key, dataset_info, self.model_format)
 
+    def active_learning(self, raw_data_location, raw_data_extension, inference_endpoint, upload_destination, conditionals):
+        '''
+        @params:
```
yeldarby, line 67: These params don't match the signature; what's _location vs _extension?

Jacobsolawetz: I had added these, just committed the new definitions

PR #27, opened 2022-04-01
NoneType concat error occurs in warning statement

ogjaylowe:


Jacobsolawetz:
Indubitably! I fixed this with better handling on our `active-learning` branch

PR #33, opened 2022-07-13
Return project IDs instead of names

yeldarby:
A user reported that they were trying to load a project from their workspace but because it returned the `name`s of the projects (which you can't do anything with) vs their `ids` which you can look up.

image

This is a quick fix to make that work more intuitively. But I think we should determine whether we actually want to return a list of `Project` objects vs strings. We have those available in this method.


FrancescoSaverioZuppichini:
@yeldarby can you rebase on master (so you have all the up to date code and the git hooks) ? :)

FrancescoSaverioZuppichini (approved):
LGTM :)

PR #34, opened 2022-07-14
passthrough api_key for sample api_keys

lukaskf:
Passthrough api_keys for samples that we use in notebooks. These will be whitelisted on the api side


yeldarby (commented):
I think we want this to be clear that they’re supposed to fill theirs in when they swap things out. Eg the string should be `YOUR-API-KEY` and only work on the one sample dataset

yeldarby (approved):

PR #48, opened 2022-09-13
enter: docker

broox:
# Description

This adds the ability to develop locally with docker.

## Type of change

-   [x] This change requires a documentation update

I was able to boot the docker container, shell into it and run tests using the instructions provided in the README.

Nope

## Docs

-   [x] Docs updated? I only updated the local `README.md`. If other documentation is required, let me know!


Jacobsolawetz (approved):
Nice! Will you be able to change code locally iteratively while in the container? 

I remember @FrancescoSaverioZuppichini showing me how to do Vscode inside of a container with the Remote Containers 

https://blog.roboflow.com/nvidia-docker-vscode-pytorch/amp/

README.md
```diff
@@ -87,14 +87,20 @@ Selecting the format you'd like your project to be exported as while choosing th
...
-5. create script on top level of `ROBOFLOW-PYTHON` and reference local `roboflow` directory: `from roboflow import Roboflow`
-6. when done, uptick the pip package minor version number in `setup.py`
-7. manually add any new dependencies to the `requirements.txt` and list of dependencies in `setup.py` (careful not to overwrite any packages that might screw up backwards dependencies for object detection, etc.) 
+## Developing locally
+
+1. clone this repo
+1. with docker
+    1. `docker build -t roboflow`
+    1. copy `.env-example` to `.env` and update environment variables for testing
+    1. shell into container: `docker run -it --rm --env-file .env -v /your/path/to/roboflow-python:/roboflow roboflow`
```
FrancescoSaverioZuppichini, line 96: I think you need to map the current user with the container

README.md
```diff
@@ -87,14 +87,20 @@ Selecting the format you'd like your project to be exported as while choosing th
...
+1. with virtualenv
+    1. create a virtual env: `virtualenv local_dev`
```
FrancescoSaverioZuppichini, line 98: You can actually do that directly from the docker container

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+VOLUME /roboflow
+WORKDIR /roboflow
+COPY README.md setup.py /roboflow/
+
+# Add GCC + CV2 dependencies
+RUN apt-get update \
+    && apt-get -y install python3-dev build-essential ffmpeg libsm6 libxext6
+
+RUN pip3 install -U pip \
```
FrancescoSaverioZuppichini, line 11: If you want to dev locally, you don't have to copy the files over, you will just install it once after you run the container

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.6-slim
```
FrancescoSaverioZuppichini, line 1:
```suggestion
FROM python:3.8-slim
```

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+VOLUME /roboflow
+WORKDIR /roboflow
+COPY README.md setup.py /roboflow/
```
FrancescoSaverioZuppichini, line 5:
```suggestion
```
No need to copy

FrancescoSaverioZuppichini (changes requested):
Thanks for working on this. I think we should address the comments and the following issues

You don't need a `Dockerfile`, if you want to have a containerized env, you can 

```
docker run -it -v $(pwd):/roboflow python:3.8.14 /bin/bash
```
Then, `cd /roboflow && pip install -e ".[dev]"`. 

So, `Dockerfile` should be removed and the doc updated. I advocate correctly.

Thanks, let me know your thoughts :)

Dockerfile
FrancescoSaverioZuppichini: 🦕

README.md
```diff
@@ -87,14 +87,20 @@ Selecting the format you'd like your project to be exported as while choosing th
...
+    1. activate virtual env: `source local_dev/bin/activate`
+    1. install packages from repo: `pip3 install -e ".[dev]"`
+1. create a script on the top level `ROBOFLOW-PYTHON` directory and reference local `roboflow` directory: `from roboflow import Roboflow`
+1. when done, uptick the pip package minor version number in `setup.py`
+1. manually add any new dependencies to the `requirements.txt` and list of dependencies in `setup.py` (careful not to overwrite any packages that might screw up backwards dependencies for object detection, etc.)
```
FrancescoSaverioZuppichini, line 103: To develop locally

### Virtual Env

Be sure to pass over the correct environments variables, see [Tests](### Tests) section

```bash
# clone repo and cd inside it
git clone git@github.com:roboflow-ai/roboflow-python.git && cd roboflow-python
# create venv
virtualenv .venv 
# activate it
source .venv/bin/activate 
# install library in edit mode
pip install -e ".[dev]"
# run test to check everything ok
python -m unittest
```

### Docker

```bash
# clone repo and cd inside it
git clone git@github.com:roboflow-ai/roboflow-python.git && cd roboflow-python
# run docker python 3.8 container and map current dir to /roboflow
docker run -it -v $(pwd):/roboflow python:3.8.14 /bin/bash
# install library in edit mode
cd /roboflow && pip install -e ".[dev]"
# run test to check everything ok
python -m unittest
```

README.md
```diff
@@ -111,12 +117,12 @@ make check_code_quality
 
 ### Tests
 
-You need to have the following `env` variables, 
+You need to have the following `env` variables defined. If using docker along with the `.env` file, these will be automatically defined.
```
FrancescoSaverioZuppichini, line 120: neat

README.md
broox: what purpose does this serve?

README.md
broox: sure, but why would you need a virtual environment inside of a containerized application?

but maybe my documentation was unclear! i was trying to give the developer 2 options. they can either use docker _or_ virtualenv.

Dockerfile
broox: i chose 3.6 here because that is the minimum supported version in `setup.py` and i thought this would be best for ensuring compatibility.

README.md
broox: 🙏 your documentation format is definitely more clear than mine.

the `Dockerfile` is certainly not necessary, but one of the reasons that i like it is because it comes with defined consistency. it also copies and installs dependencies upon initial build.

with the suggested approach of not using a `Dockerfile`, every time you start the container, you have to manually CD into the project's root directory and reinstall dependencies before you can use the application.

i also chose `python:3.6-slim` because `3.6` is the minimum version supported in `setup.py` and the `slim` image is almost 8 times smaller than the `python:3.8.14` image...

i'm curious about your thoughts here!

Dockerfile
broox: this is true, but then you have to install dependencies every time you start the container. i personally like to automate this upon build to save time while developing later.

README.md
FrancescoSaverioZuppichini: to ensure everything written inside the container can be accessed by the current host user. E.g. maybe I am logging something to a file, once I close the container I won't be able to edit it, since it was created by a different user

Dockerfile
FrancescoSaverioZuppichini: Replied on slack :)

FrancescoSaverioZuppichini (approved):
LGTM! Thanks a lot @broox

PR #53, opened 2022-09-16
Added Dockerfile.dev

FrancescoSaverioZuppichini:
# Description

Created `Dockerfile.dev` to allow us to quickly spin up a docker container with all the requirements deps to dev. Follow up to #48 

```bash
docker build -t roboflow-python -f Dockerfile.dev .
# or with a specific python version
docker build  -t roboflow-python --build-arg PYTHON_VERSION=3.9 -f Dockerfile.dev .
docker run --rm -it \
  -v $(pwd)/:/worspace/ \
  roboflow-python 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,16 @@ docker run --rm -it \
 python -m unittest
 ```
 
+#### Change Python version
+
+You can pass the build arg `PYTHON_VERSION` to dynamically change python version at build time
+
+```
+docker build  -t roboflow-python --build-arg PYTHON_VERSION=3.9 -f Dockerfile.dev .
+```
+
+Wil use `python:3.9-slim`
+
 **Note** If you are using [VSCode](https://code.visualstudio.com/) we reccomend to read the ["Developing inside a Container"](https://code.visualstudio.com/docs/remote/containers) tutorial.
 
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -101,23 +101,32 @@ git clone git@github.com:roboflow-ai/roboflow-python.git && cd roboflow-python
...
-# Build and run a new docker container image
-docker run -it -v $(pwd):/roboflow --name roboflow python:3.8.14-slim /bin/bash
+# Build our development image
+docker build -t roboflow-python -f Dockerfile.dev .
 
-# Install Dependencies
-cd roboflow/ && pip install -e ".[dev]"
+# Run container and map current folder in it
+docker run --rm -it \
+  -e .env
```
broox, line 109: after the mocking PR is merged, i think that we will have no need for `.env` or `.env-example`. those values were only used in tests and are no longer necessary 😄

FrancescoSaverioZuppichini: nice catch! Thanks. Updated :)

README.md
```diff
@@ -101,23 +101,31 @@ git clone git@github.com:roboflow-ai/roboflow-python.git && cd roboflow-python
 # Be sure to update the values with your account's information
 cp .env-example .env
```
broox, line 102:
```suggestion
```

we can remove this and the comment above it.

we can also remove the `.env-example` file and everything about the `.env` variables from the `Testing` section below.

paulguerrie (approved):
Looks good!

PR #54, opened 2022-09-20
Define model before using it in the README

broox:
# Description

The example code in the README doesn't define `model` before using it.


FrancescoSaverioZuppichini (approved):
LGTM! Thanks :)

PR #55, opened 2022-09-20
Enter: Instance segmentation

broox:
# Description

This adds Instance Segmentation functionality to the python pip package. I also did a small bit of refactoring but tried to avoid going wild.

## Type of change

A couple unit tests have been added. I also tested the new functionality manually via:

```
import os
import roboflow

ROBOFLOW_API_KEY = os.environ.get("ROBOFLOW_API_KEY")
WORKSPACE_NAME = os.environ.get("WORKSPACE_NAME", "")
PROJECT_NAME = os.environ.get("PROJECT_NAME")
PROJECT_VERSION = os.environ.get("PROJECT_VERSION", "1")

LOCAL_IMAGE = "<local image path>"
HOSTED_IMAGE = "<remote image url>"

# Instantiate Roboflow object with your API key
rf = roboflow.Roboflow(api_key=ROBOFLOW_API_KEY)

# List all projects for your workspace
workspace = rf.workspace()
print(f'workspace: {workspace}')

# Load a certain project, workspace url is optional
project = rf.project(PROJECT_NAME)
print(f'project: {project}')

# List all versions of a specific project
project.versions()
print(f'versions: {project.versions()}')

print(f'images: {project.images}')

# Upload image to dataset
# project.upload(LOCAL_IMAGE)

# Retrieve the model of a specific project
model = project.version(PROJECT_VERSION).model
print(f'model: {model}')

# # predict on a local image
# prediction = model.predict(LOCAL_IMAGE)
# print(prediction)

# Predict on a hosted image
prediction = model.predict(HOSTED_IMAGE) # , hosted=True)
print(prediction)

# Plot the prediction
prediction.plot()
print(prediction.plot())

# Convert predictions to JSON
prediction.json()
print(prediction.json())

# Save the prediction as an image
prediction.save(output_path='predictions.jpg')
```

Once this is merged, we will need to update the docs that are displayed on instance segmentation projects.

eg:

### Infer on Local and Hosted Images

To install dependencies, `pip install roboflow`

````
from roboflow import Roboflow

rf = Roboflow(api_key="API_KEY")
project = rf.workspace().project("MODEL_ENDPOINT")
model = project.version(VERSION).model

# infer on a local image
print(model.predict("your_image.jpg").json())

# infer on an image hosted elsewhere
print(model.predict("https://example.com/your_image.jpg").json())

# change the confidence for your prediction
print(model.predict("your_image.jpg", confidence=60).json())

# visualize your prediction
model.predict("your_image.jpg").save("prediction.jpg")
````

## Docs

### DIFF
--- a/roboflow/config.py
+++ b/roboflow/config.py
@@ -1,13 +1,23 @@
 import os
 
+CLASSIFICATION_MODEL = os.getenv("CLASSIFICATION_MODEL", default="ClassificationModel")
+INSTANCE_SEGMENTATION_MODEL = "InstanceSegmentationModel"
 OBJECT_DETECTION_MODEL = os.getenv(
     "OBJECT_DETECTION_MODEL", default="ObjectDetectionModel"
 )
-CLASSIFICATION_MODEL = os.getenv("CLASSIFICATION_MODEL", default="ClassificationModel")
 PREDICTION_OBJECT = os.getenv("PREDICTION_OBJECT", default="Prediction")
+
 API_URL = os.getenv("API_URL", default="https://api.roboflow.com")
 APP_URL = os.getenv("APP_URL", default="https://app.roboflow.com")
+INSTANCE_SEGMENTATION_URL = os.getenv(
+    "INSTANCE_SEGMENTATION_URL", "https://outline.roboflow.com"
+)
+
 CLIP_FEATURIZE_URL = os.getenv(
     "CLIP_FEATURIZE_URL", default="CLIP FEATURIZE URL NOT IN ENV"
 )
 DEMO_KEYS = ["coco-128-sample", "chess-sample-only-api-key"]
+
+TYPE_CLASSICATION = "classification"
+TYPE_OBJECT_DETECTION = "object-detection"
+TYPE_INSTANCE_SEGMENTATION = "instance-segmentation"
--- a/roboflow/models/instance_segmentation.py
+++ b/roboflow/models/instance_segmentation.py
@@ -29,22 +29,28 @@ class InstanceSegmentationModel:
 
         self.api_url = f"{INSTANCE_SEGMENTATION_URL}/{dataset_id}/{version}"
 
-    def predict(self, image_path):
+    def predict(self, image_path, confidence=40):
         """
         Infers detections based on image from specified model and image path
 
-        :param image_path: Path to image (can be local or hosted)
-        :return: PredictionGroup --> a group of predictions based on Roboflow JSON response
+        :param image_path: Path to image (can be local path or hosted URL)
+        :param confidence: A threshold for the returned predictions on a scale of 0-100. A lower number will return more predictions. A higher number will return fewer, high-certainty predictions.
+
+        :return: PredictionGroup - a group of predictions based on Roboflow JSON response
         """
         validate_image_path(image_path)
 
-        params = {"api_key": self.__api_key}
+        params = {
+            "api_key": self.__api_key,
+            "confidence": confidence,
+        }
         request_kwargs = {}
 
         hosted_image = urllib.parse.urlparse(image_path).scheme in (
             "http",
             "https",
         )
+
         if hosted_image:
             params["image"] = image_path
         else:
@@ -61,6 +67,7 @@ class InstanceSegmentationModel:
 
         url = f"{self.api_url}?{urllib.parse.urlencode(params)}"
         response = requests.post(url, **request_kwargs)
+        response.raise_for_status()
 
         return PredictionGroup.create_prediction_group(
             response.json(),
--- a/roboflow/models/instance_segmentation.py
+++ b/roboflow/models/instance_segmentation.py
@@ -1,0 +1,72 @@
+import io
+import urllib
+
+import requests
+from PIL import Image
+from requests_toolbelt.multipart.encoder import MultipartEncoder
+
+from roboflow.config import INSTANCE_SEGMENTATION_MODEL, INSTANCE_SEGMENTATION_URL
+from roboflow.util.image_utils import validate_image_path
+from roboflow.util.prediction import PredictionGroup
+
+
+class InstanceSegmentationModel:
+    def __init__(
+        self,
+        api_key,
+        version_id,
+    ):
+        """
+        :param api_key: Your API key (obtained via your workspace API settings page)
+        :param version_id: The ID of the dataset version to use for predicting
+        """
+        self.__api_key = api_key
+        self.id = version_id
+
+        version_info = self.id.rsplit("/")
+        dataset_id = version_info[1]
+        version = version_info[2]
+
+        self.api_url = f"{INSTANCE_SEGMENTATION_URL}/{dataset_id}/{version}"
+
+    def predict(self, image_path):
+        """
+        Infers detections based on image from specified model and image path
+
+        :param image_path: Path to image (can be local or hosted)
+        :return: PredictionGroup --> a group of predictions based on Roboflow JSON response
+        """
+        validate_image_path(image_path)
+
+        params = {"api_key": self.__api_key}
+        request_kwargs = {}
+
+        hosted_image = urllib.parse.urlparse(image_path).scheme in (
+            "http",
+            "https",
+        )
+        if hosted_image:
+            params["image"] = image_path
+        else:
+            image = Image.open(image_path)
+            buffered = io.BytesIO()
+            image.save(buffered, quality=90, format="JPEG")
+            data = MultipartEncoder(
+                fields={"file": ("imageToUpload", buffered.getvalue(), "image/jpeg")}
+            )
+            request_kwargs = {
+                "data": data,
+                "headers": {"Content-Type": data.content_type},
+            }
+
+        url = f"{self.api_url}?{urllib.parse.urlencode(params)}"
+        response = requests.post(url, **request_kwargs)
+
+        return PredictionGroup.create_prediction_group(
+            response.json(),
+            image_path=image_path,
+            prediction_type=INSTANCE_SEGMENTATION_MODEL,
+        )
+
+    def __str__(self):
+        return f"<{type(self).__name__} id={self.id}, api_url={self.api_url}>"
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: roboflow/models/instance_segmentation.py [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+class InstanceSegmentationModel:
+    def __init__(
+        self,
+        api_key,
+        version_id,
+    ):
+        """
+        :param api_key: Your API key (obtained via your workspace API settings page)
+        :param version_id: The ID of the dataset version to use for predicting
+        """
```
broox, line 22: A couple of the existing model classes seemed to be a bit over-engineered - at least with my current understanding in how these are being used.

I tried to simplify this model as much as possible by making it only using the data that I think it needs.

e.g. The dataset and version can be obtained from the `version_id`, so that is the only argument I require on `init()`. The `local` argument on the other models seemed like a URL string that could be passed in to override the base URL for local/staging development... it seemed a bit cleaner to me to just use `.env` variables for this, so that's the route I went here. If we like this, maybe the others can be similarly refactored.

I also didn't understand the purpose of the `load_model()` method on the other models... nor did I see it used anywhere in the codebase, so I did not add it here.

I nearly made a new `Inference` class that we could make each of these child classes extend from. But that seemed like a rabbit hole best saved for another day. If this is missing anything, please let me know!

roboflow/models/instance_segmentation.py
```diff
@@ -0,0 +1,72 @@
...
```
broox, line 22: A couple of the existing model classes seemed to be a bit over-engineered - at least with my current understanding in how these are being used.

I tried to simplify this model as much as possible by making it only using the data that I think it needs.

e.g. The dataset and version can be obtained from the `version_id`, so that is the only argument I require on `init()`. The `local` argument on the other models seemed like a URL string that could be passed in to override the base URL for local/staging development... it seemed a bit cleaner to me to just use `.env` variables for this, so that's the route I went here. If we like this, maybe the others can be similarly refactored.

I also didn't understand the purpose of the `load_model()` method on the other models... nor did I see it used anywhere in the codebase, so I did not add it here.

I nearly made a new `Inference` class that we could make each of these child classes extend from. But that seemed like a rabbit hole best saved for another day. If this is missing anything, please let me know!

roboflow/models/instance_segmentation.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self.api_url = f"{INSTANCE_SEGMENTATION_URL}/{dataset_id}/{version}"
+
+    def predict(self, image_path):
+        """
+        Infers detections based on image from specified model and image path
+
+        :param image_path: Path to image (can be local or hosted)
+        :return: PredictionGroup --> a group of predictions based on Roboflow JSON response
+        """
```
broox, line 38: The `predict()` method on the other models accepted a `hosted` flag that seemed a little redundant. We can determine if something is hosted or not by looking at the `image_path`, so I omitted it here. We can maybe deprecate this argument on the other models as well.

roboflow/util/image_utils.py
```diff
@@ -1,10 +1,40 @@
...
-    return False
+def validate_image_path(image_path):
+    """
+    Validate whether a local OR hosted image path is valid
+    :param image_path: local path or URL of image
+    :returns: None
+    :raises Exception: Image path is not valid
+    """
+    if not check_image_path(image_path):
+        raise Exception(f"Image does not exist at {image_path}!")
```
broox, line 40: I noticed a lot of code duplication in the codebase for validating an `image_path`, so I made a couple generic helpers for this. We could refactor much of the codebase to use these helpers in the future.

roboflow/util/prediction.py
```diff
@@ -83,6 +84,15 @@ def plot_annotation(axes, prediction=None, stroke=1):
...
         )
+    elif prediction["prediction_type"] == INSTANCE_SEGMENTATION_MODEL:
+        points = [[p["x"], p["y"]] for p in prediction["points"]]
+        polygon = patches.Polygon(
+            points,
+            linewidth=stroke,
+            edgecolor="r",
+            facecolor="none",
+        )
+        axes.add_patch(polygon)
```
broox, line 95: i was not exactly sure how to test this method, but this _seems_ right?

roboflow/util/prediction.py
```diff
@@ -367,6 +389,17 @@ def save(self, output_path="predictions.jpg", stroke=2):
...
+            elif self.base_prediction_type == INSTANCE_SEGMENTATION_MODEL:
+                points = [[int(p["x"]), int(p["y"])] for p in prediction["points"]]
+                np_points = np.array(points, dtype=np.int32)
+                cv2.polylines(
+                    image,
+                    [np_points],
+                    isClosed=True,
+                    color=stroke_color,
+                    thickness=stroke,
+                )
```
broox, line 401: There's a lot of duplicated image logic between `Prediction` and `PredictionGroup`. We should refactor this to reuse the same logic between the two classes.

roboflow/config.py
```diff
@@ -1,13 +1,23 @@
 import os
 
+CLASSIFICATION_MODEL = os.getenv("CLASSIFICATION_MODEL", default="ClassificationModel")
+INSTANCE_SEGMENTATION_MODEL = "InstanceSegmentationModel"
```
broox, line 4: i didn't allow this to be overridden from the `env` because i didn't understand the use-case. i'm happy to make that change if it makes sense tho.

roboflow/models/instance_segmentation.py
```diff
@@ -0,0 +1,72 @@
...
+        request_kwargs = {}
+
+        hosted_image = urllib.parse.urlparse(image_path).scheme in (
+            "http",
+            "https",
+        )
+        if hosted_image:
+            params["image"] = image_path
+        else:
+            image = Image.open(image_path)
```
broox, line 51: other models seemed to open with `cv2` and then convert to `PIL`... i didn't understand that use-case, so i just used `PIL` here. if this makes sense we can refactor the others accordingly.

roboflow/models/instance_segmentation.py
```diff
@@ -0,0 +1,72 @@
...
+        response = requests.post(url, **request_kwargs)
+
+        return PredictionGroup.create_prediction_group(
+            response.json(),
+            image_path=image_path,
+            prediction_type=INSTANCE_SEGMENTATION_MODEL,
+        )
+
+    def __str__(self):
+        return f"<{type(self).__name__} id={self.id}, api_url={self.api_url}>"
```
broox, line 72: other models returned formatted `JSON` here, which was a bit foreign to me. I went with a string approach that seems to be more common, but I'm happy to change this if y'all prefer the JSON.

roboflow/util/prediction.py
```diff
@@ -450,26 +483,19 @@ def create_prediction_group(json_response, image_path, prediction_type):
...
                 )
-                # Add to prediction list
                 prediction_list.append(prediction)
             img_dims = json_response["image"]
-        # For classification model
         elif prediction_type == CLASSIFICATION_MODEL:
-            # Create prediction for predicted class
             prediction = Prediction(json_response, image_path, prediction_type)
-            # add to prediction list
             prediction_list.append(prediction)
```
broox, line 497: these comments all felt so redundant. the code is self explanatory.

roboflow/models/instance_segmentation.py
broox, line 22: A couple of the existing model classes seemed to be a bit over-engineered - at least with my current understanding in how these are being used.

I tried to simplify this model as much as possible by making it only using the data that I think it needs.

e.g. The dataset and version can be obtained from the `version_id`, so that is the only argument I require on `init`. The `local` argument on the other models seemed like a URL string that could be passed in to override the base URL for local/staging development... it felt a bit cleaner to me to just use `.env` variables for this, so that's the route I went here. If we like this, maybe the others can be similarly refactored.

I also didn't understand the purpose of the `load_model()` method on the other models... nor did I see it used anywhere in the codebase, so I did not add it here.

I nearly made a new `Inference` class that we could make each of these child classes extend from. But that seemed like a rabbit hole best saved for another day. If this is missing anything, please let me know!

broox:
@paulguerrie interesting! can you help me find an example API response of that? I'd love to test it out.

roboflow/models/instance_segmentation.py
```diff
@@ -0,0 +1,72 @@
...
+        self.__api_key = api_key
+        self.id = version_id
+
+        version_info = self.id.rsplit("/")
+        dataset_id = version_info[1]
+        version = version_info[2]
+
+        self.api_url = f"{INSTANCE_SEGMENTATION_URL}/{dataset_id}/{version}"
+
+    def predict(self, image_path):
```
Jacobsolawetz, line 32: Let's add confidence as a param as well, should flow through like the object detection confidence param does

roboflow/models/instance_segmentation.py
Jacobsolawetz: Simplifying looks good to me! local model exists for local deployments when someone spins up a local inference docker container, so they could hit that local server - it doesn't exist for inst seg yet so let's leave it ablated!

roboflow/models/instance_segmentation.py
Jacobsolawetz: Nice

roboflow/util/prediction.py
Jacobsolawetz: Seems good on vis on my side

roboflow/util/prediction.py
Jacobsolawetz: Definitely

Jacobsolawetz:
@paulguerrie bump on if you have an example model + im for that edge case?

Jacobsolawetz (approved):
Looks great! I added some comments to the threads you opened and suggested adding confidence (which may be the only code change)

When we're ready let's add the snippets throughout! https://github.com/roboflow-ai/roboflow/pull/1371 - I'm game to pair on that core app code too if you want

And perhaps we can weight for sem seg and knock those out together

PR #60, opened 2022-10-04
Fixes to Image Upload Routes

Jacobsolawetz:
# Description

While testing out proxying PIP uploads in [this PR](https://github.com/roboflow-ai/license-server/pull/5), I found a couple of issues to be addressed - proposals in code

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -235,7 +235,7 @@ class Project:
             except UnidentifiedImageError:
                 return False
         else:
-            #for hosted images check to see if the extension is in our list of valid extensions
+            # for hosted images check to see if the extension is in our list of valid extensions
             if any(ext in image_path for ext in ACCEPTED_IMAGE_EXTENSIONS):
                 valid = True
             else:
@@ -325,9 +325,7 @@ class Project:
         # User gives image path
         if image_path is not None:
             # Upload Image Response
-            response = self.__image_upload(
-                image_path, is_hosted=is_hosted, split=split
-            )
+            response = self.__image_upload(image_path, is_hosted=is_hosted, split=split)
             # Get JSON response values
             try:
                 if "duplicate" in response.json().keys():
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -14,6 +14,7 @@ from roboflow.config import API_URL, DEMO_KEYS
 from roboflow.core.version import Version
 
 ACCEPTED_IMAGE_FORMATS = ["PNG", "JPEG"]
+ACCEPTED_IMAGE_EXTENSIONS = [".jpg", ".jpeg", ".png"]
 
 
 def custom_formatwarning(msg, *args, **kwargs):
@@ -135,15 +136,15 @@ class Project:
 
         raise RuntimeError("Version number {} is not found.".format(version_number))
 
-    def __image_upload(self, image_path, hosted_image=False, split="train"):
+    def __image_upload(self, image_path, is_hosted=False, split="train"):
         """function to upload image to the specific project
         :param image_path: path to image you'd like to upload.
-        :param hosted_image: if the image is hosted online, then this should be modified
+        :param is_hosted: if the image is hosted online, then this should be modified
         :param split: the dataset split to upload the project to.
         """
 
         # If image is not a hosted image
-        if not hosted_image:
+        if not is_hosted:
 
             image_name = os.path.basename(image_path)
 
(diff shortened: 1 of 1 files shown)


roboflow/core/project.py
```diff
@@ -145,14 +145,13 @@ def __image_upload(self, image_path, hosted_image=False, split="train"):
         # If image is not a hosted image
         if not hosted_image:
 
-            project_name = self.id.rsplit("/")[1]
             image_name = os.path.basename(image_path)
 
             # Construct URL for local image upload
             self.image_upload_url = "".join(
                 [
-                    "https://api.roboflow.com/dataset/",
```
Jacobsolawetz, line 154: In some cases we may want to change the API_URL from ENV, this seemed to have stuck around as hard coded

roboflow/core/project.py
```diff
@@ -270,7 +271,10 @@ def upload(
             )
 
         if is_file:
-            is_image = self.check_valid_image(image_path) or is_hosted
+            if is_hosted:
```
Jacobsolawetz, line 274: hit bug here that check_valid_image returns error when checking if `is_hosted`

Jacobsolawetz: Yeah hmm - I'm going to go with:

* take is_hosted
* check the extension of image_path

The download and check maybe not a huge deal, but seems to defeat purpose of having everything happen up in the cloud - ideally infer server will handle good error there :D

roboflow/core/project.py
```diff
@@ -259,6 +257,9 @@ def upload(
             "https://"
         )
 
+        if not hosted_image:
```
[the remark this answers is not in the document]
Jacobsolawetz: Roger - updating to use a unified `is_hosted` - certainly seems cleaner

roboflow/core/project.py
```diff
@@ -227,21 +226,27 @@ def __annotation_upload(self, annotation_path, image_id):
...
+        if not is_hosted:
+            try:
+                img = Image.open(image_path)
+                valid = img.format in ACCEPTED_IMAGE_FORMATS
+                img.close()
+            except UnidentifiedImageError:
+                return False
+        else:
+            # for hosted images check to see if the extension is in our list of valid extensions
+            if any(ext in image_path for ext in ACCEPTED_IMAGE_EXTENSIONS):
```
broox, line 239: this would allow a file named `jpg.exe` to pass and is also case insensitive.
i believe this should solve both of those issues:

```suggestion
            if any(image_path.lower().endswith(ext) for ext in ACCEPTED_IMAGE_EXTENSIONS):
```

it'd be nice to add some unit tests for this function to ensure that it's catching the right things... this logic could also be helpful in `image_utils` as a new function, like `check_image_format(image_path)`.

roboflow/core/project.py
```diff
@@ -227,21 +226,27 @@ def __annotation_upload(self, annotation_path, image_id):
...
+                valid = True
```
broox, line 240: this could just `return True`, but this works too

roboflow/core/project.py
```diff
@@ -250,7 +255,7 @@ def upload(
         """upload function
         :param image_path: path to image you'd like to upload
         :param annotation_path: if you're upload annotation, path to it
-        :param hosted_image: whether the image is hosted
+        :param is_hosted: whether the image is hosted
```
broox, line 258: we can kill this docstring cause `is_hosted` gets defined at the beginning of this function rather than being pulled from the function `kwargs`.

broox: i almost wonder if we should create a helper method to determine this in `util/image_utils.py` or a new utility file, like `util/file_utils.py`... that would allow the logic to be easily reusable, updatable, testable, and also make it so that we don't have to pass this kwarg around. e.g:

```
def is_hosted(path):
    return image_path.startswith("http://") or image_path.startswith(
        "https://"
    )
```

just a thought.

roboflow/core/project.py
```diff
@@ -227,21 +226,27 @@ def __annotation_upload(self, annotation_path, image_id):
...
-        except UnidentifiedImageError:
-            return False
+    def check_valid_image(self, image_path, is_hosted=False):
+        if not is_hosted:
+            try:
+                img = Image.open(image_path)
+                valid = img.format in ACCEPTED_IMAGE_FORMATS
+                img.close()
+            except UnidentifiedImageError:
+                return False
```
FrancescoSaverioZuppichini, line 236:
```suggestion
image = Image.open(image_path).convert("RGB")
```
If the image is invalid, PIL will raise an Exception that is perfectly fine since the user will be notified

roboflow/core/project.py
FrancescoSaverioZuppichini: Same as before, no need to hardcode formats here imho

roboflow/core/project.py
FrancescoSaverioZuppichini: better way imho

```
is_hosted = not Path(<image_path).exists()

```

FrancescoSaverioZuppichini (commented):
Added comments

FrancescoSaverioZuppichini:
@Jacobsolawetz what's the status of this PR?

Jacobsolawetz:
Closing since various fixes herein were applied by @ogjaylowe and @hansent in other PRs

PR #61, opened 2022-10-05
Login from CLI or Python

Jacobsolawetz:
# Description

This is a draft PR for logging in with the CLI or natively in Python

It will remain a draft until CLI auth solidifies

### DIFF
--- a/roboflow/config.py
+++ b/roboflow/config.py
@@ -58,8 +58,8 @@ if RF_WORKSPACE == None:
     RF_API_KEY = None
 else:
     RF_API_KEY = None
-    for k in WORKPACES.keys():
-        workspace = WORKPACES[k]
+    for k in RF_WORKPACES.keys():
+        workspace = RF_WORKPACES[k]
         if workspace["url"] == RF_WORKSPACE:
             RF_API_KEY = workspace["apiKey"]
 # ENV API_KEY OVERRIDE
--- a/roboflow/config.py
+++ b/roboflow/config.py
@@ -1,19 +1,48 @@
 import os
+import json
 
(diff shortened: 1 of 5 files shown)


roboflow/config.py
```diff
@@ -1,23 +1,67 @@
...
+
+
+def get_var_conditional(key, default):
+
+    if os.getenv(key) != None:
+        return os.getenv(key)
+    elif key in config.keys():
+        return config[key]
+    else:
+        return default
```
broox, line 25: could we avoid the complexity of 2 config files by reading the response of the login json and then just setting the relevant ENV variable(s) (rather than writing it to a file)? the downside here would be that each new terminal instance would require a new login. are there security concerns with leaving the user's API keys on the filesystem?

roboflow/config.py
```diff
@@ -1,23 +1,67 @@
...
+RF_WORKSPACE = get_var_conditional("RF_WORKSPACE", default=None)
+# DEFAULT_WORKSPACE = get_var_conditional("default_workspace", default=None)
+if RF_WORKSPACE == None:
+    RF_API_KEY = None
+else:
+    RF_API_KEY = None
+    for k in RF_WORKPACES.keys():
+        workspace = RF_WORKPACES[k]
+        if workspace["url"] == RF_WORKSPACE:
+            RF_API_KEY = workspace["apiKey"]
```
broox, line 64: could do a `break` here to avoid looping after one is found.

broox (commented):
left a few nitpicky suggestions and questions!

Jacobsolawetz:
Closing since this was absorbed in quickstart backbones

PR #63, opened 2022-10-06
Add simple version.export() functionality and clean up version.download() a bit

broox:
# Description

While adding `Version.export()`, and thinking about how that could be integrated into `Version.download()`, I decided to also clean that method up a bit to make it more parseable and testable. I think I also fixed a couple bugs in the process (and hopefully didn't introduce any 😅)

## Type of change

I've only tested with the new unit tests so far. We should definitely manually test. I'd love to pair on that.

## Docs

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -109,7 +109,7 @@ class Version:
             link = "https://app.roboflow.com/ds/n9QwXwUK42?key=NnVCe2yMxP"
         else:
             url = self.__get_download_url(model_format)
-            resp = requests.get(url, params={'api_key': self.__api_key })
+            resp = requests.get(url, params={"api_key": self.__api_key})
             if resp.status_code == 200:
                 link = resp.json()["export"]["link"]
             else:
@@ -134,7 +134,7 @@ class Version:
         :raises RuntimeError / HTTPError:
         """
         url = self.__get_download_url(model_format)
-        response = requests.post(url, params={'api_key': self.__api_key })
+        response = requests.post(url, params={"api_key": self.__api_key})
         if not response.ok:
             try:
                 raise RuntimeError(response.json())
@@ -153,6 +153,7 @@ class Version:
 
         :return None:
         """
+
         def bar_progress(current, total, width=80):
             progress_message = (
                 "Downloading Dataset Version Zip in "
@@ -186,7 +187,7 @@ class Version:
                 try:
                     zip_ref.extract(member, location)
                 except zipfile.error:
-                    raise RuntimeError('Error unzipping download')
+                    raise RuntimeError("Error unzipping download")
 
         os.remove(location + "/roboflow.zip")
 
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -43,6 +43,10 @@ class Version:
         else:
             self.__api_key = api_key
             self.name = name
+
+            # FIXME: the version argument is inconsistently passed into this object.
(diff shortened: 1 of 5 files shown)


roboflow/core/version.py
```diff
@@ -53,7 +57,7 @@ def __init__(self, version_dict, type, api_key, name, version, model_format, loc
             self.splits = version_dict["splits"]
             self.model_format = model_format
 
-            version_without_workspace = os.path.basename(version)
+            version_without_workspace = os.path.basename(str(version))
```
broox, line 60: this was failing if `version` was a number... which seems to be a thing that happens.

roboflow/core/version.py
```diff
@@ -85,54 +89,77 @@ def __init__(self, version_dict, type, api_key, name, version, model_format, loc
...
-                location = (
-                    os.environ["DATASET_DIRECTORY"]
-                    + "/"
-                    + self.name.replace(" ", "-")
-                    + "-"
-                    + self.version
-                )
-            else:
-                location = self.name.replace(" ", "-") + "-" + self.version
+            location = self.__get_download_location()
```
broox, line 101: this same URL is helpful for extracting to, so i popped it into a private helper.

roboflow/core/version.py
```diff
@@ -85,54 +89,77 @@ def __init__(self, version_dict, type, api_key, name, version, model_format, loc
...
-                RuntimeError(
-                    "You must pass a download_type to version.download() or define model in your Roboflow object"
-                )
-
-        if model_format == "yolov5":
-            model_format = "yolov5pytorch"
-
-        if model_format == "yolov7":
-            model_format = "yolov7pytorch"
+        model_format = self.__get_format_identifier(model_format)
```
broox, line 106: i tried to simplify/DRY up the `model_format` logic above and also fixed the issue in that the `RuntimeError` was never being `raised`. i believe that my added unit tests should cover all of the cases.

roboflow/core/version.py
FrancescoSaverioZuppichini: nice catch!

roboflow/core/version.py
FrancescoSaverioZuppichini: Out of curiosity, do you know why `__` are used all around?

roboflow/core/version.py
```diff
@@ -85,54 +89,77 @@ def __init__(self, version_dict, type, api_key, name, version, model_format, loc
...
+        :return: True
+        :raises RuntimeError / HTTPError:
+        """
+        url = self.__get_download_url(model_format)
+        response = requests.post(url, params={"api_key": self.__api_key})
+        if not response.ok:
+            try:
+                raise RuntimeError(response.json())
+            except requests.exceptions.JSONDecodeError:
+                response.raise_for_status()
```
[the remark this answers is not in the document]
FrancescoSaverioZuppichini: neat

FrancescoSaverioZuppichini (approved):
LGTM! 🚀 💜

roboflow/core/version.py
broox: good question!

this is a pattern that i was following in the library where "private" class methods seemed to be prefixed by a double underscore.

usually, in python a single underscore is used to define a private method, but either work. there's slightly more protection with the double underscore because outside of `self`, the method name is mangled a bit and harder to call directly.

FrancescoSaverioZuppichini: Thanks, the double `__` looks weird imho, no strong feelings as well

PR #68, opened 2022-10-13
Clip compare

ogjaylowe:
# Description

Given a `dir` of images and a `target` image, perform CLIP encoding and similarity comparison using our hosted CLIP backend (uses cosine similarity method) and returns an object mapping each image's similarity score from the `dir` to the `target` image. 

No required dependencies that weren't already in there. I added `twine` to the list of `dev` packages as we use twine to update the hosted package up in the pypi store.

You must specify the following URL in `config.py` as the env variable `CLIP_FEATURIZE_URL`, `'https://clip-featurize-double-image-li37nwjfaq-uc.a.run.app/predict'`, to hit the endpoint as this endpoint does not have auth / usage tracking.

## Type of change

I had a directory of images and a target image that I ran the script against. Checked for the following edge cases and added `assert` statements as needed`:
- no image or directory string provided (including `None` type)
- image or directory provided but doesn't exist (handled by built-in pillow exceptions)

YOUR_ANSWER

Can't think of any.

## Docs

-   [ ] Docs updated? What were the changes: will not update docs until endpoint gets auth / usage tracking. Instead will be providing Walmart the URL directly with instructions on how to use the pip package accordingly (as per Brad's suggestion).

### DIFF
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -11,7 +11,6 @@ from roboflow.util.active_learning_utils import (
     clip_encode,
     count_comparisons,
 )
-
 from roboflow.util.clip_compare_utils import clip_encode
 
 
@@ -82,7 +81,7 @@ class Workspace:
             dir: (str) = name reference to a directory of images for comparison
             image_ext: (str) = file format for expected images (don't include the . before the file type name)
             target_image: (str) = name reference for target image to compare individual images from directory against
-                
+
             returns: (dict) = a key:value mapping of image_name:comparison_score_to_target
         """
 
@@ -95,15 +94,14 @@ class Workspace:
 
         # grab all images in a given directory with ext type
         for image in glob.glob(f"./{dir}/*{image_ext}"):
-            # compare image 
+            # compare image
             similarity = clip_encode(image, target_image)
 
             # map image name to similarity score
-            comparisons.append({image:similarity})
+            comparisons.append({image: similarity})
 
         # TODO - sort all comparisons to find best matches?
         return comparisons
-
 
     def active_learning(
         self,
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -78,7 +78,7 @@ class Workspace:
 
     def clip_compare(
         self, dir: str = "", image_ext: str = ".png", target_image: str = ""
-    ) -> dict[str, float]:
+    ) -> dict:
         """
         @params:
             dir: (str) = name reference to a directory of images for comparison
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -97,7 +97,6 @@ def clip_compare(self, dir: str = "", image_ext: str = ".png", target_image: str
             similarity = clip_encode(image, target_image)
             # map image name to similarity score
             comparisons.append({image: similarity})
-
         # TODO - sort all comparisons to find best matches?
         return comparisons
 
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -12,6 +12,8 @@ from roboflow.util.active_learning_utils import (
     count_comparisons,
 )
 
+from roboflow.util.clip_compare_utils import clip_encode
+
 
 class Workspace:
     def __init__(self, info, api_key, default_workspace, model_format):
@@ -73,6 +75,35 @@ class Workspace:
         dataset_info = dataset_info.json()["project"]
 
         return Project(self.__api_key, dataset_info, self.model_format)
+
+    def clip_compare(self, dir="", image_ext=".png", target_image=""):
+        """
+        @params:
+            dir: (str) = name reference to a directory of images for comparison
+            image_ext: (str) = file format for expected images (don't include the . before the file type name)
+            target_image: (str) = name reference for target image to compare individual images from directory against
+                
+            returns: (dict) = a key:value mapping of image_name:comparison_score_to_target
+        """
+
+        assert isinstance(dir, str), "valid directory string required!"
+        assert isinstance(image_ext, str), "valid image ext string required!"
+        assert isinstance(target_image, str), "valid target image string required!"
+
+        # list to store comparison results in
+        comparisons = []
+
+        # grab all images in a given directory with ext type
+        for image in glob.glob(f"./{dir}/*{image_ext}"):
+            # compare image 
+            similarity = clip_encode(image, target_image)
+
+            # map image name to similarity score
+            comparisons.append({image:similarity})
+
+        # TODO - sort all comparisons to find best matches?
+        return comparisons
+
 
     def active_learning(
         self,
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -1,6 +1,7 @@
 import glob
 import json
 import sys
+from PIL import Image
 
 import requests
 
@@ -75,7 +76,9 @@ class Workspace:
 
         return Project(self.__api_key, dataset_info, self.model_format)
 
-def clip_compare(self, dir: str = "", image_ext: str = ".png", target_image: str = "") -> Dict[str, float]:
+    def clip_compare(
+        self, dir: str = "", image_ext: str = ".png", target_image: str = ""
+    ) -> dict[str, float]:
         """
         @params:
             dir: (str) = name reference to a directory of images for comparison
@@ -85,10 +88,6 @@ def clip_compare(self, dir: str = "", image_ext: str = ".png", target_image: str
             returns: (dict) = a key:value mapping of image_name:comparison_score_to_target
         """
 
-        assert isinstance(dir, str), "valid directory string required!"
-        assert isinstance(image_ext, str), "valid image ext string required!"
-        assert isinstance(target_image, str), "valid target image string required!"
-
         # list to store comparison results in
         comparisons = []
         # grab all images in a given directory with ext type
@@ -97,7 +96,7 @@ def clip_compare(self, dir: str = "", image_ext: str = ".png", target_image: str
             similarity = clip_encode(image, target_image)
             # map image name to similarity score
             comparisons.append({image: similarity})
-comparisons = sorted(comparisons, key=lambda item: -list(item.values())[0])
+            comparisons = sorted(comparisons, key=lambda item: -list(item.values())[0])
         return comparisons
 
     def active_learning(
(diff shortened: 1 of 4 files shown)


roboflow/util/clip_compare_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    @params:
+        iamge_path: (str) = name reference to a given image for encoding
+
+    returns: a base64 encoded string formatted for travel to an HTTP endpoint
+    """
+
+    image = Image.open(image_path)
+    buffered = io.BytesIO()
+    image_rgb = image.convert("RGB")
+    image_rgb.save(buffered, quality=90, format="JPEG")
```
FrancescoSaverioZuppichini, line 24: why JPEG with a lower quality?

roboflow/util/clip_compare_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    img_str = base64.b64encode(buffered.getvalue())
+    return img_str.decode("ascii")
+
+
+# base64 encode two images and send them to the clip endpoint for further encoding and comparison
+def clip_encode(image1, image2):
```
FrancescoSaverioZuppichini, line 30:
```suggestion
def clip_encode(image1: str, image2: str) -> float:
```

roboflow/util/clip_compare_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    image1 = base64_encode(image1)
+    image2 = base64_encode(image2)
+
+    url = CLIP_FEATURIZE_URL
+
+    headers = {"Content-Type": "text/plain"}
+    data = json.dumps({"image1": image1, "image2": image2})
+
+    r = requests.post(url, data=data, headers=headers)
```
FrancescoSaverioZuppichini, line 46:
```suggestion
r = requests.post(CLIP_FEATURIZE_URL, data=data, headers=headers)
```

roboflow/util/clip_compare_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    @params:
+        image1: (str) = name referenceto a given image for encoding
+        image2: (str) = name referenceto a given image for encoding
+
+        returns: (float) = a value between 1 and 0, with 1 being images were identical
+    """
+    image1 = base64_encode(image1)
+    image2 = base64_encode(image2)
+
+    url = CLIP_FEATURIZE_URL
```
FrancescoSaverioZuppichini, line 41:
```suggestion
```

setup.py
```diff
@@ -38,7 +38,7 @@
     ],
     packages=find_packages(exclude=("tests",)),
     extras_require={
-        "dev": ["flake8", "black==22.3.0", "isort", "responses"],
+        "dev": ["flake8", "black==22.3.0", "isort", "responses","twine"],
```
FrancescoSaverioZuppichini, line 41: what is `twine`?

FrancescoSaverioZuppichini (commented):
Thanks! 💜  Added some stile comments. 

I am not sure a formatted was run, we have black there you can run it via `black <path_to_your_files>`. 

I am not sure about the docstring style, looks like javadoc that is not used normally. I advocate for google style docstring

roboflow/core/workspace.py
```diff
@@ -74,6 +75,34 @@ def project(self, project_name):
 
         return Project(self.__api_key, dataset_info, self.model_format)
 
+    def clip_compare(self, dir="", image_ext=".png", target_image=""):
```
FrancescoSaverioZuppichini, line 78:
```suggestion
def clip_compare(self, dir: str = "", image_ext: str = ".png", target_image: str = "") -> Dict[str, float]:
```

roboflow/core/workspace.py
```diff
@@ -74,6 +75,34 @@ def project(self, project_name):
...
+        assert isinstance(target_image, str), "valid target image string required!"
+
+        # list to store comparison results in
+        comparisons = []
+
+        # grab all images in a given directory with ext type
+        for image in glob.glob(f"./{dir}/*{image_ext}"):
+            # compare image
+            similarity = clip_encode(image, target_image)
+
```
FrancescoSaverioZuppichini, line 99:
```suggestion
````

File: roboflow/core/workspace.py [hunk 8]
```diff
@@ -74,6 +75,34 @@ def project(self, project_name):
...
+            returns: (dict) = a key:value mapping of image_name:comparison_score_to_target
+        """
+
+        assert isinstance(dir, str), "valid directory string required!"
+        assert isinstance(image_ext, str), "valid image ext string required!"
+        assert isinstance(target_image, str), "valid target image string required!"
+
+        # list to store comparison results in
+        comparisons = []
+
```
FrancescoSaverioZuppichini, line 94:
```suggestion
````

File: roboflow/core/workspace.py [hunk 9]
```diff
@@ -74,6 +75,34 @@ def project(self, project_name):
...
+        """
+        @params:
+            dir: (str) = name reference to a directory of images for comparison
+            image_ext: (str) = file format for expected images (don't include the . before the file type name)
+            target_image: (str) = name reference for target image to compare individual images from directory against
+
+            returns: (dict) = a key:value mapping of image_name:comparison_score_to_target
+        """
+
+        assert isinstance(dir, str), "valid directory string required!"
```
FrancescoSaverioZuppichini, line 88: raise `ValueError` instead, `assert` is bad practice

roboflow/core/workspace.py
```diff
@@ -74,6 +75,34 @@ def project(self, project_name):
...
+        comparisons = []
+
+        # grab all images in a given directory with ext type
+        for image in glob.glob(f"./{dir}/*{image_ext}"):
+            # compare image
+            similarity = clip_encode(image, target_image)
+
+            # map image name to similarity score
+            comparisons.append({image: similarity})
+
```
FrancescoSaverioZuppichini, line 102:
```suggestion
```

roboflow/core/workspace.py
```diff
@@ -74,6 +75,34 @@ def project(self, project_name):
...
+        # TODO - sort all comparisons to find best matches?
```
FrancescoSaverioZuppichini, line 103:
```suggestion
comparisons = sorted(comparisons, key=lambda item: -list(item.values())[0])
```
The data representation is weird, you should have something like

```
[ 
   { "name" : str
    "similarity: float}, 
       { "name" : str
    "similarity: float},
   ...
]

Not sure why you decided to have dynamic key names (depending on image name)

File: roboflow/util/clip_compare_utils.py [hunk 12]
```diff
@@ -0,0 +1,48 @@
...
+import io
+import json
+
+import requests
+from PIL import Image
+
+from roboflow.config import CLIP_FEATURIZE_URL
+
+
+# rf.predict requires images formatted and base64 encoded
```
FrancescoSaverioZuppichini, line 12:
```suggestion
```

roboflow/util/clip_compare_utils.py
```diff
@@ -0,0 +1,48 @@
...
+def base64_encode(image_path):
+    """
```
FrancescoSaverioZuppichini, line 14:
```suggestion
""" `rf.predict` requires images formatted and base64 encoded

File: roboflow/util/clip_compare_utils.py [hunk 14]
```diff
@@ -0,0 +1,48 @@
...
+
+    image = Image.open(image_path)
+    buffered = io.BytesIO()
+    image_rgb = image.convert("RGB")
+    image_rgb.save(buffered, quality=90, format="JPEG")
+    img_str = base64.b64encode(buffered.getvalue())
+    return img_str.decode("ascii")
+
+
+# base64 encode two images and send them to the clip endpoint for further encoding and comparison
```
FrancescoSaverioZuppichini, line 29:
```suggestion
```

roboflow/util/clip_compare_utils.py
```diff
@@ -0,0 +1,48 @@
...
+def clip_encode(image1, image2):
+    """
+    @params:
```
FrancescoSaverioZuppichini, line 32:
```suggestion
""" This function base64 encode two images and send them to the clip endpoint for further encoding and comparison

File: setup.py (hunk 5 above)
Comment (FrancescoSaverioZuppichini) [reply setup.py:41]: oh got it :) I didn't know it about it, cool!

File: roboflow/core/workspace.py (hunk 11 above)
Comment (ogjaylowe) [reply roboflow/core/workspace.py:100]: Nairwita (developer on the Walmart team) and I white boarded the original resulting obj to be queried by filename so you could take random samples from a range easier--hence the dynamic key names for O(1) lookup.

Since you put in the sorting command I think we'd still be able to accomplish this with your recommended code format and use the resulting list indexes as a way to query instead. We don't really care about how we look things up so long as we can sample things like `top 10 matches`, `10 matches randomly sampled from the 70th to 90th percentile of comparisons`, etc.

File: roboflow/util/clip_compare_utils.py (hunk 1 above)
Comment (ogjaylowe) [reply roboflow/util/clip_compare_utils.py:23]: I ripped this out of some sample code from deeper in the code base and didn't consider changing the `format`. What should we be using as a standard? I think `.tiff` files technically have the best quality but not sure if we should be using something more common such as `.png`.

File: setup.py (hunk 5 above)
Comment (ogjaylowe) [reply setup.py:41]: Yes we use `twine` when publishing updates as documented here: https://roboflow.slab.com/posts/publishing-roboflow-pip-packages-k1a2z4ck

File: roboflow/core/workspace.py (hunk 9 above)
Comment (ogjaylowe) [reply roboflow/core/workspace.py:88]: I have removed the `assert` statements. Since `AttributeError` would be the main cause for concern here I decided to remove error handling and allow `PIL` to do it for us.

**Comment by ogjaylowe** [comment]
FYI I used the `make style` command to execute Makefile we had in the repo before the push that ran `black` for me. I'm not sure if it's supposed to be doing the docstring formatting or not.

I have added some additional info about proper docstring formats following the [google docstring format](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for future contributions.

File: roboflow/util/clip_compare_utils.py (hunk 1 above)
Comment (FrancescoSaverioZuppichini) [reply roboflow/util/clip_compare_utils.py:23]: If we don't know why this is done, let's just store it as png with untouched quality

File: roboflow/core/workspace.py [hunk 16]
```diff
@@ -78,7 +78,7 @@ def project(self, project_name):
 
     def clip_compare(
         self, dir: str = "", image_ext: str = ".png", target_image: str = ""
-    ) -> dict[str, float]:
+    ) -> dict:
```
FrancescoSaverioZuppichini, line 81:
```suggestion
    ) -> Dict[str, float]:
```
Does it work?

FrancescoSaverioZuppichini (approved):
LGTM! I've left some comments

PR #69, opened 2022-10-13
OpenCV Image Encoder

Tyler-Odenthal:
# Description

We added cv2 to our package list because we want to use it for image encoding. This allows us to bypass saving an image locally and might save inference time.

I added some logic to fix when an end user tried to send an OpenCV frame array. Previously the logic would crash and we provided a statement to ensure that doesn't happen for frame array data.

## List any dependencies that are required for this change.

import cv2

Tested on multiple webcam and resolutions.

Still passes images paths and image URLs just fine.

Not sure the size of the cv2 package, but it is something to consider. Probably worth storage space if it speeds up inference logic.

## Docs

-   [ ] Docs updated? 

What were the changes: Docs have currently not been updated, but I will when change goes into effect. Will add example of using OpenCV2 frame data in model.predict()

### DIFF
--- a/roboflow/models/object_detection.py
+++ b/roboflow/models/object_detection.py
@@ -147,7 +147,7 @@ class ObjectDetectionModel:
 
                 # Create buffer
                 buffered = io.BytesIO()
-                image.save(buffered, quality=90, format="JPEG")
+                image.save(buffered, format="PNG")
                 # Base64 encode image
                 img_str = base64.b64encode(buffered.getvalue())
                 img_str = img_str.decode("ascii")
--- a/roboflow/models/object_detection.py
+++ b/roboflow/models/object_detection.py
@@ -3,6 +3,7 @@ import io
 import json
 import os
 import urllib
+import cv2
 
 import requests
 from PIL import Image
@@ -118,7 +119,7 @@ class ObjectDetectionModel:
         """
         Infers detections based on image from specified model and image path
 
-        :param image_path: Path to image (can be local or hosted)
+        :param image_path: Path to image or image array (can be local or hosted)
         :param hosted: If image located on a hosted server, hosted should be True
         :param format: output format from this method
         :return: PredictionGroup --> a group of predictions based on Roboflow JSON response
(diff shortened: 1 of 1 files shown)


roboflow/models/object_detection.py
```diff
@@ -132,27 +133,42 @@ def predict(
...
-            img_str = base64.b64encode(buffered.getvalue())
-            img_str = img_str.decode("ascii")
-
-            # Post to API and return response
-            resp = requests.post(
-                self.api_url,
-                data=img_str,
-                headers={"Content-Type": "application/x-www-form-urlencoded"},
-            )
+            if ".jpg" in image_path or ".png" in image_path:  # Open Image in RGB Format
```
FrancescoSaverioZuppichini, line 145: should we support other formats as well?

roboflow/models/object_detection.py
```diff
@@ -132,27 +133,42 @@ def predict(
...
+                image = Image.open(image_path).convert("RGB")
+
+                # Create buffer
+                buffered = io.BytesIO()
+                image.save(buffered, quality=90, format="JPEG")
```
FrancescoSaverioZuppichini, line 150: why JPEG and 90 quality?

FrancescoSaverioZuppichini (approved):
LGTM

roboflow/models/object_detection.py
FrancescoSaverioZuppichini: discussed here as well -> https://github.com/roboflow-ai/roboflow-python/pull/68 (in the comments)

roboflow/models/object_detection.py
FrancescoSaverioZuppichini: discussed here as well -> https://github.com/roboflow-ai/roboflow-python/pull/68 (in the comments) 👀

roboflow/models/object_detection.py
Tyler-Odenthal: Feel free to support other image formats, those were just the ones that have been tested.

Using imencode saves time when you are using OpenCV2 to read data. How we currently have the package coded will not accept OpenCV2 frames because there is no logic to handle Numpy arrays.

By using the image encoder this way we ensure that no one need to write an image on either side of the deployment when using OpenCV2.

Crashing

In the scenario above, I will need to write an image to work with the predict function. I would rather pass the cv2 frame array.

roboflow/models/object_detection.py
Tyler-Odenthal: I removed the quality flag and changed the format to "PNG"

It was legacy from another script.

PR #70, opened 2022-10-21
upticking minor version number for pypi deploy

ogjaylowe:
need to uptick minor version number for latest merges to go live into pypi


broox (approved):
i dunno, this looks pretty sketchy

PR #73, opened 2022-10-27
Two stage detection method

ogjaylowe:
# Description

Users can specify their two model names and versions to do two stage inference.

No new dependencies.

## Type of change

Tested for different model types and various parameter validation (`None` type, not default type, etc.)

## Docs

-   [x] Docs updated? Will update docs with new method param upon merge.

### DIFF
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -1,7 +1,7 @@
 import glob
 import json
-import sys
 import os
+import sys
 
 import requests
 from PIL import Image
@@ -134,7 +134,10 @@ class Workspace:
         # perform first inference
         predictions = stage_one_model.predict(image)
 
-        if stage_one_project.type == "object-detection" and stage_two_project == "classification":
+        if (
+            stage_one_project.type == "object-detection"
+            and stage_two_project == "classification"
+        ):
             # interact with each detected object from stage one inference results
             for boundingbox in predictions:
                 # rip bounding box coordinates from json1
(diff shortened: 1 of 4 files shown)


roboflow/core/workspace.py
```diff
@@ -99,6 +100,116 @@ def clip_compare(
...
+                    boundingbox["x"] + boundingbox["width"] / 2,
+                    boundingbox["y"] + boundingbox["height"] / 2,
+                )
+
+                # create a new cropped image using the first stage prediction coordinates (for each box!)
+                croppedImg = pil_image.crop(box)
+                croppedImg.save("./temp.png")
+
+                # capture results of second stage inference from cropped image
+                results.append(stage_two_model.predict("./temp.png")[0])
```
Jacobsolawetz, line 154: Could clean up temp.png after

roboflow/core/workspace.py
```diff
@@ -99,6 +100,116 @@ def clip_compare(
...
+        elif stage_one_project.type == "classification":
+            print(
+                "please use an object detection model--can only perform two stage with bounding box results"
+            )
+
+        return results
```
Jacobsolawetz, line 160: Anything else we want to do for post proc of the results? I.E. scale them back up to the dimensions of the first image?

If this is for classification only - add a model2 restrictor?

roboflow/core/workspace.py
```diff
@@ -99,6 +100,116 @@ def clip_compare(
...
+                croppedImg = pil_image.crop(box)
+
+                # capture OCR results from cropped image
+                results.append(ocr_infer(croppedImg)["results"])
+        else:
+            print(
+                "please use an object detection model--can only perform two stage with bounding box results"
+            )
+
+        return results
```
Jacobsolawetz, line 211: Similar to above, anything we want to do on post processing the results? Should it keep the bounding box info, for example?

Jacobsolawetz (approved):

PR #76, opened 2022-11-16
Local active learning

ogjaylowe:
# Description

Updated `Version` so users can specify a locally running inference server within the pip package.

## Type of change

Test code:
```
from roboflow import Roboflow

rf = Roboflow(api_key="BLlkFnwfSaRFUXBfU0tJ")

model_name = "hard-hat-universe-0dy7t"
model_number = 17
upload_img_directory = "./"
image_extention_type = ".png"
project_upload_destation = "merge_tester"

conditionals = {
    "required_objects_count" : 1,
    "required_class_count": 1,
    "target_classes": [],
    "minimum_size_requirement" : float('-inf'),
    "maximum_size_requirement" : float('inf'),
    "confidence_interval" : [10,90],
}

rf.workspace().active_learning(upload_img_directory, image_extention_type, [model_name, model_number], project_upload_destation, conditionals, True)
```

Tested edge cases:
- provide `True` and inference server is up
- provide `True` and inference server is down
- provide `False` and inference server is up

Tested with Cobravision and oak devices in mind but tested on several servers.

## Docs

-   [ ] Docs updated? What were the changes: Will update docs for active learning and general inference via local server when merged.

### DIFF
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -280,12 +280,13 @@ class Workspace:
         )
 
         # check if inference_model references endpoint or local
-        if use_localhost == True:
-            inference_model = f"http://localhost:9001/{inference_endpoint[0]}/{inference_endpoint[1]}?api_key={self.__api_key}"
-        else:
-            inference_model = (
-                self.project(inference_endpoint[0]).version(inference_endpoint[1]).model
-            )
+        local = "http://localhost:9001/" if use_localhost else None
+
+        inference_model = (
+            self.project(inference_endpoint[0])
+            .version(version_number=inference_endpoint[1], local=local)
+            .model
+        )
         upload_project = self.project(upload_destination)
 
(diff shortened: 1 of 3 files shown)


roboflow/core/workspace.py
```diff
@@ -277,8 +279,13 @@ def active_learning(
             else conditionals["maximum_size_requirement"]
         )
 
+        # check if inference_model references endpoint or local
+        local = "http://localhost:9001/" if use_localhost else None
```
ogjaylowe, line 283: food for thought, we could wrap up the `"http://localhost:9001/"` in the `version` object if we don't want to give users the option to specify alternative IP addresses or ports.

I know Dwayne will sometimes call a direct IP instead of the localhost but not sure what would be best.

Jacobsolawetz: Ok for now

PR #79, opened 2022-12-01
Remove and ignore .DS_Store.

ClaytonJY:
# Description

Remove the `.DS_Store` file, and add it to `.gitignore` to prevent it from coming back.

This is a MacOS-specific file that doesn't need to be here.


Jacobsolawetz (approved):
Yes, indeed! Thanks @ClaytonJY

PR #81, opened 2022-12-01
Retry Key Validation in Sparse Server Failure Case

Jacobsolawetz:
# Description

Retry auth key validation in the event of a sparsely occurring server authentication failure

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -1,5 +1,6 @@
 import json
 import sys
+import time
 
 import requests
 
@@ -10,7 +11,7 @@ from roboflow.core.workspace import Workspace
 __version__ = "0.2.20"
 
 
-def check_key(api_key, model, notebook):
+def check_key(api_key, model, notebook, num_retries=0):
     if type(api_key) is not str:
         raise RuntimeError(
             "API Key is of Incorrect Type \n Expected Type: "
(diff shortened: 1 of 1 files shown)


roboflow/__init__.py
```diff
@@ -28,11 +29,23 @@ def check_key(api_key, model, notebook):
...
+                    print("retrying...")
+                    time.sleep(1)
+                    num_retries += 1
+                    return check_key(api_key, model, notebook, num_retries)
+                else:
+                    raise RuntimeError(
+                        "There was an error validating the api key with Roboflow server."
+                    )
             else:
+                r = response.json()
```
yeldarby, line 48: Should we put a try/catch around this & retry if it fails to produce valid JSON (but wit ha 200 error code) too?

yeldarby: We should prob do this retry logic around all API routes not just the initialization

Jacobsolawetz: Thought about that... I think this is safe if we get a 200 code, it will always produce valid JSON

Jacobsolawetz: Interesting on _all_ routes - before undertaking that, do we think all api routes have this same vulnerability?

FrancescoSaverioZuppichini (approved):
LGTM! :rocket:

PR #90, opened 2022-12-28
Rstp stream support

ogjaylowe:
# Description

updated input handling to support cv2 ndarrays so that an RSTP stream could send frames via VideoCapture() directly to the active_learning() function.

## Type of change


Tyler-Odenthal (approved):
Looks good on my end! Should be able to accept cv2 data and append results.

PR #93, opened 2023-01-06
Fix for json_response["image"] in prediction.py

Tyler-Odenthal:
# Description

There was an issue because we changed our data structure recently. The resulted in json_response["image"] not existing.

Used PIL to get image dimensions and reconstructed img_dims.

_No dependency changes_

## Type of change

Tested on two TRT scripts and with three different models. Have yet to test it on an edge device, but this fix should not be device specific.

Reading image for image dimensions takes time, if there is a more efficient way to get img_dims or if we are already pulling dims somewhere else in the script. We should pass the variable instead of creating another variable.

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -8,7 +8,7 @@ from roboflow.config import API_URL, APP_URL, DEMO_KEYS
 from roboflow.core.project import Project
 from roboflow.core.workspace import Workspace
 
-__version__ = "0.2.23"
+__version__ = "0.2.24"
 
 
 def check_key(api_key, model, notebook, num_retries=0):
--- a/roboflow/util/prediction.py
+++ b/roboflow/util/prediction.py
@@ -488,8 +488,10 @@ class PredictionGroup:
         """
         prediction_list = []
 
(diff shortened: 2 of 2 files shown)


roboflow/__init__.py
```diff
@@ -8,7 +8,7 @@
 from roboflow.core.project import Project
 from roboflow.core.workspace import Workspace
 
-__version__ = "0.2.23"
+__version__ = "0.2.24"
```
FrancescoSaverioZuppichini, line 11: why the version was changed? Are you planning on a new release? It's better to make a single separate PR with the version bump to avoid overlapping with other PRs

FrancescoSaverioZuppichini (approved):
LGTM, thanks @Tyler-Odenthal 

For the future, don't change the version number - **only change it in a self contained PR**

roboflow/__init__.py
Tyler-Odenthal: Noted! Appreciate the feedback! 

Is this still good to push or do I need to revert the version number?

FrancescoSaverioZuppichini: Now you have to push it since we increase the number from another PR :)

PR #96, opened 2023-01-12
add dataset download overwrite flag

SkalskiP:
# Description

This change adds a dataset download overwrite flag. By default, it is set to `True` to keep the current default behavior. When set to `False,` the dataset won't get downloaded if it is already available locally. 

The motivation for that change is my current work on [PR](https://github.com/ultralytics/ultralytics/pull/295) to YOLOv8. The code there is structured so that it would re-download the dataset multiple times.

## Type of change

I can change the docs if we already have documentation for that method somwhare.


Jacobsolawetz (approved):
Makes sense!

PR #97, opened 2023-01-23
Fix Ultralytics Upload Parse

Jacobsolawetz:
# Description

Ultralytics updated their model object variable types, breaking our external model deploy pathway. In this PR, we fix those issues.


paulguerrie (approved):
LGTM after deving and testing with Jacob

PR #98, opened 2023-01-24
Support Old Versions of Ultralytics YOLOv8 Deploy

Jacobsolawetz:
# Description

Support old versions of Ultralytics YOLOv8 deploy - they changed the class types of their model args, but we may still want to parse those older models

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -284,7 +284,7 @@ class Version:
 
         return True
 
-    def deploy(self, model_type: str, model_path: str) -> None:
+   def deploy(self, model_type: str, model_path: str) -> None:
         """Uploads provided weights file to Roboflow
 
         Args:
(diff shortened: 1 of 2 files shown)


roboflow/core/version.py
```diff
@@ -318,18 +318,32 @@ def deploy(self, model_type: str, model_path: str) -> None:
...
+                    "nc": model["model"].nc,
+                    "args": {
+                        k: val
+                        for k, val in model["model"].args.items()
+                        if ((k == "model") or (k == "imgsz") or (k == "batch"))
+                    },
+                    "ultralytics_version": ultralytics.__version__,
+                    "model_type": model_type,
+                }
+            except:
```
bigbitbus, line 334: if we knew the exception thrown then some other error will not get caught and handled with this catch, and you could have a "catch all" for that case

bigbitbus (approved):
Since we dont know the exact error, the catch all except is the best option.

Jacobsolawetz:
cc @paulguerrie

PR #102, opened 2023-01-26
[PoC] Warning for specific versions

FrancescoSaverioZuppichini:
# Description

This PR includes a series of functions inside `util.versions.py` to check and print warning when a dependency with a version we don't support is installed in the env. 

> **Note**  We are reading from `requirements.txt` the required packages and placing them into `setup.py` for consistency

```python
from roboflow.util.versions import print_warn_for_wrong_dependencies_versions

print_warn_for_wrong_dependencies_versions([("torch", "<=", "0.3.0")])
```

prints

```
Dependency torch<=0.3.0 is required but found version=1.13.1, to fix: `pip install torch<=0.3.0`
```

There is also a decorator

```python
@warn_for_wrong_dependencies_versions([("torch", "==", "1.2.0")])
def foo(x):
    # I only work with torch `1.2.0` but another one is installed
    print(f"foo {x}")
```

prints:

```
Dependency torch==1.2.0 is required but found version=1.13.1, to fix: `pip install torch==1.2.0`
```

# Motivation

@SkalskiP told me we are reaching a tons of messages due to our sdk not working with `ultralytics>=8.11.0`, this will generate a warning and hopefully help the users


Jacobsolawetz (approved):
This will be a nice quality of life improvement @SkalskiP @FrancescoSaverioZuppichini let's remember to bump this along with notebooks as we pull our support up

PR #105, opened 2023-01-27
Silently fail to print the warning in version.download if Ultralytics not installed

FrancescoSaverioZuppichini:
# Description

If `ultralytics` is not installed in `version.download`, the `print_warn_for_wrong_dependencies_versions` will throw an `ImportError`, since there are no `ultralytics` modules.

We now catch it and `pass` without printing anything. This makes sense because if the user tries to use `.deploy` later, we will inform him he needs to have `ultralytics` installed. 

I could also print a warning instead of `pass` it, down for suggestions

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -166,7 +166,9 @@ class Version:
                     [("ultralytics", "<=", "8.0.20")]
                 )
             except ImportError as e:
-                print("[WARNING] we noticed you are downloading a `yolov8` datasets but you don't have `ultralytics` installed. Roboflow `.deploy` supports only models trained with `ultralytics<=8.0.20`, to intall it `pip install ultralytics<=8.0.20`.")
+                print(
+                    "[WARNING] we noticed you are downloading a `yolov8` datasets but you don't have `ultralytics` installed. Roboflow `.deploy` supports only models trained with `ultralytics<=8.0.20`, to intall it `pip install ultralytics<=8.0.20`."
+                )
                 # silently fail
                 pass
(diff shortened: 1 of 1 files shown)


roboflow/core/version.py
```diff
@@ -158,10 +159,15 @@ def download(self, model_format=None, location=None, overwrite: bool = True):
...
-            )
+            # if ultralytics is installed, we will assume users will want to use yolov8 and we check for the supported version
+            try:
+                import_module("ultralytics")
+                print_warn_for_wrong_dependencies_versions(
+                    [("ultralytics", "<=", "8.0.20")]
+                )
+            except ImportError as e:
+                # silently fail
+                pass
```
SkalskiP, line 170: We could keep it as it is. Or add some info about our ultralytics support.

FrancescoSaverioZuppichini: Happy to do both, since you have been exposed to the final user more than me, what do you think we should do?

FrancescoSaverioZuppichini: Copy that

SkalskiP: You will need to install it to call `.deploy` later, you can use: `pip install ultralytics<=8.0.20`.

to

Roboflow `.deploy` supports only models trained with `ultralytics<=8.0.20`

Thoughts?

FrancescoSaverioZuppichini: I've merged the two messages :angel:

Jacobsolawetz (approved):

PR #109, opened 2023-02-01
Quickstart Backbones

Jacobsolawetz:
# Description

This PR adds bones to the python package to support a streamlined quickstart experience:

Noteably a:

* .login()
* .load_model()
* .webcam()

## Docs

- yes - in quickstart

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,3 +18,4 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 wget
 requests_toolbelt
+ipython>=8.9.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,4 +18,3 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 wget
 requests_toolbelt
-ipython>=8.9.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,4 +18,4 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 wget
 requests_toolbelt
-ipython>=8.9.1
+ipython>=8.9.0
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -1,7 +1,7 @@
 import json
+import os
 import sys
 import time
-import os
 
 import requests
 
@@ -11,6 +11,7 @@ from roboflow.core.workspace import Workspace
 from roboflow.util.general import write_line
 
 __version__ = "0.2.29"
+
 
 def check_key(api_key, model, notebook, num_retries=0):
     if type(api_key) is not str:
@@ -49,7 +50,7 @@ def check_key(api_key, model, notebook, num_retries=0):
                 r = response.json()
                 return r
     else:  # then you're using a dummy key
-        
+
         sys.stdout.write(
             "upload and label your dataset, and get an API KEY here: "
             + APP_URL
@@ -61,21 +62,27 @@ def check_key(api_key, model, notebook, num_retries=0):
         )
         return "onboarding"
 
+
 def auth(api_key):
     r = check_key(api_key)
     w = r["workspace"]
 
     return Roboflow(api_key, w)
 
+
 def login(force=False):
-    
+
     conf_location = os.getenv(
-        "ROBOFLOW_CONFIG_DIR", default=os.getenv("HOME") + "/.config/roboflow/config.json"
+        "ROBOFLOW_CONFIG_DIR",
+        default=os.getenv("HOME") + "/.config/roboflow/config.json",
     )
-    
+
     if os.path.isfile(conf_location) and not force:
-        raise(Exception("You are already logged in. To re-login, run roboflow.login(force=True)"))
-    
+        write_line(
+            "You are already logged into Roboflow. To make a different login, run roboflow.login(force=True)."
+        )
+        return
+
     write_line("visit " + APP_URL + "/auth-cli" " to get your authentication token.")
 
     token = input("Paste the authentication here token here: ")
@@ -85,55 +92,58 @@ def login(force=False):
     if r_login.status_code == 200:
 
         r_login = r_login.json()
-        
-        #make config directory if it doesn't exist
+
+        # make config directory if it doesn't exist
         if not os.path.exists(os.path.dirname(conf_location)):
             os.mkdir(os.path.dirname(conf_location))
-            
+
         r_login = {"workspaces": r_login}
-        #set first workspace as default workspace
-        
+        # set first workspace as default workspace
+
         default_workspace_id = list(r_login["workspaces"].keys())[0]
         workspace = r_login["workspaces"][default_workspace_id]
         r_login["RF_WORKSPACE"] = workspace["url"]
-        
-        #write config file
+
+        # write config file
         with open(conf_location, "w") as f:
             json.dump(r_login, f, indent=2)
-        
+
     else:
         raise RuntimeError("Error logging in")
 
 
-#import roboflow
-#roboflow.#action
+# import roboflow
+# roboflow.#action
 
 active_workspace = None
 
+
 def initialize_roboflow():
-    
+
     global active_workspace
-    
+
     conf_location = os.getenv(
-        "ROBOFLOW_CONFIG_DIR", default=os.getenv("HOME") + "/.config/roboflow/config.json"
+        "ROBOFLOW_CONFIG_DIR",
+        default=os.getenv("HOME") + "/.config/roboflow/config.json",
     )
-    
+
     if not os.path.isfile(conf_location):
-        raise("To use this method, you must first login - run roboflow.login()")
+        raise ("To use this method, you must first login - run roboflow.login()")
     else:
         if active_workspace == None:
-            active_workspace = Roboflow().workspace()   
-        
-        return active_workspace       
-    
+            active_workspace = Roboflow().workspace()
+
+        return active_workspace
+
+
 def train():
-    
+
     operate_workspace = initialize_roboflow()
-    
-    #check if project or version is set
-    
+
+    # check if project or version is set
+
     print("training from " + operate_workspace.name)
-    
+
 
 def load_model(project=None, version=None):
     operate_workspace = initialize_roboflow()
@@ -143,17 +153,16 @@ def load_model(project=None, version=None):
     return model
 
 
-#def set_project()
+# def set_project()
 
 # def set_version()
-    
-#download()
 
-#deploy()
+# download()
+
+# deploy()
 
 
-
-#continue distributing this object for back compatibility
+# continue distributing this object for back compatibility
 class Roboflow:
     def __init__(
         self,
@@ -164,7 +173,7 @@ class Roboflow:
         self.api_key = api_key
         if self.api_key == None:
             self.api_key = GET_RF_API_KEY()
-        
+
         self.model_format = model_format
         self.notebook = notebook
         self.onboarding = False
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -50,7 +50,6 @@ def check_key(api_key, model, notebook, num_retries=0):
                 r = response.json()
                 return r
     else:  # then you're using a dummy key
-
         sys.stdout.write(
             "upload and label your dataset, and get an API KEY here: "
             + APP_URL
@@ -71,7 +70,6 @@ def auth(api_key):
 
 
 def login(force=False):
-
     conf_location = os.getenv(
         "ROBOFLOW_CONFIG_DIR",
         default=os.getenv("HOME") + "/.config/roboflow/config.json",
@@ -90,7 +88,6 @@ def login(force=False):
     r_login = requests.get(APP_URL + "/query/cliAuthToken/" + token)
 
     if r_login.status_code == 200:
-
         r_login = r_login.json()
 
         # make config directory if it doesn't exist
@@ -119,7 +116,6 @@ active_workspace = None
 
 
 def initialize_roboflow():
-
     global active_workspace
 
     conf_location = os.getenv(
@@ -137,7 +133,6 @@ def initialize_roboflow():
 
 
 def train():
-
     operate_workspace = initialize_roboflow()
 
     # check if project or version is set
@@ -145,7 +140,7 @@ def train():
     print("training from " + operate_workspace.name)
 
 
-def load_model(project=None, version=None):
+def load_model(project=None, version=None, local=None):
     operate_workspace = initialize_roboflow()
     project = operate_workspace.project(project)
     version = project.version(version)
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -5,7 +5,7 @@ import os
 
 import requests
 
-from roboflow.config import API_URL, APP_URL, DEMO_KEYS
+from roboflow.config import API_URL, APP_URL, DEMO_KEYS, GET_RF_API_KEY
 from roboflow.core.project import Project
 from roboflow.core.workspace import Workspace
 from roboflow.util.general import write_line
@@ -49,6 +49,8 @@ def check_key(api_key, model, notebook, num_retries=0):
                 r = response.json()
                 return r
     else:  # then you're using a dummy key
+        
+        print("API_KEY: " + api_key)
         sys.stdout.write(
             "upload and label your dataset, and get an API KEY here: "
             + APP_URL
@@ -89,6 +91,7 @@ def login():
         #set first workspace as default workspace
         
         default_workspace_id = list(r_login["workspaces"].keys())[0]
+        workspace = r_login["workspaces"][default_workspace_id]
         r_login["RF_WORKSPACE"] = workspace["url"]
         
         #write config file
@@ -101,11 +104,14 @@ def login():
 class Roboflow:
     def __init__(
         self,
-        api_key="YOUR ROBOFLOW API KEY HERE",
+        api_key=None,
         model_format="undefined",
         notebook="undefined",
     ):
         self.api_key = api_key
+        if self.api_key == None:
+            self.api_key = GET_RF_API_KEY()
+        
         self.model_format = model_format
         self.notebook = notebook
         self.onboarding = False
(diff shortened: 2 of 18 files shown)


Jacobsolawetz:
```
#!pip install jupyterlab
#!pip install ipywidgets
#!jupyter labextension install @jupyter-widgets/jupyterlab-manager
#!jupyter nbextension enable --py widgetsnbextension --sys-prefix
#!jupyter nbextension enable --user --py widgetsnbextension  --sys-prefix
```

roboflow/config.py
```diff
@@ -1,26 +1,58 @@
...
+else:
+    config = {}
+
+
+# read configuration variables conditionally
+
+
+##1. check if variable is in environment
+##2. check if variable is in config file
+##3. return default value
```
FrancescoSaverioZuppichini, line 22: Please, use [python docstring in google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

roboflow/config.py
```diff
@@ -1,26 +1,58 @@
...
+def get_var_conditional(key, default):
+    if os.getenv(key) != None:
+        return os.getenv(key)
+    elif key in config.keys():
+        return config[key]
+    else:
+        return default
```
FrancescoSaverioZuppichini, line 29: Please, add a comment or use a better variables name to make it easier.

For instance, what's the type of `default`?

FrancescoSaverioZuppichini: for instance, I am not sure but the above code may be rewritten by

`os.environ.get(key, config.get(key, default))`

roboflow/config.py
```diff
@@ -30,3 +62,23 @@
...
+        RF_API_KEY = None
+        for k in RF_WORKSPACES.keys():
+            workspace = RF_WORKSPACES[k]
+            if workspace["url"] == RF_WORKSPACE:
+                RF_API_KEY = workspace["apiKey"]
+    # ENV API_KEY OVERRIDE
+    if os.getenv("RF_API_KEY") != None:
+        RF_API_KEY = os.getenv("RF_API_KEY")
+
+    return RF_API_KEY
```
FrancescoSaverioZuppichini, line 84: Note that general good naming practices dictate that only `GLOBAL_VARIABLES` should be named in `UPPER_CASE`, in this case this is a function that get's exectute everytime this module is imported

roboflow/core/version.py
```diff
@@ -344,11 +365,14 @@ def deploy(self, model_type: str, model_path: str) -> None:
...
-        class_names = []
-        for i, val in enumerate(model["model"].names):
-            class_names.append((val, model["model"].names[val]))
-        class_names.sort(key=lambda x: x[0])
-        class_names = [x[1] for x in class_names]
+        if isinstance(model["model"].names, list):
+            class_names = model["model"].names
+        else:
+            class_names = []
+            for i, val in enumerate(model["model"].names):
```
FrancescoSaverioZuppichini, line 372: `enumerate` works on an iterable, what should be the type expected here? Since we checked for `list` before

roboflow/models/classification.py
```diff
@@ -13,7 +13,16 @@
...
+    def __init__(
+        self,
+        api_key,
+        id,
+        name=None,
+        version=None,
+        local=False,
+        colors={},
+        preprocessing={},
+    ):
```
FrancescoSaverioZuppichini, line 25: This is wrong, you shouldn't use mutable classes as parameters, set them to `None` and create it if is `None`

`colors=None`

Then in the `__init__`

`self.colors = {} if colors is None else colors`

roboflow/models/classification.py
```diff
@@ -60,13 +72,19 @@ def predict(self, image_path, hosted=False):
             # Create API URL for hosted image (slightly different)
             self.api_url += "&image=" + urllib.parse.quote_plus(image_path)
             # POST to the API
-            resp = requests.get(self.api_url)
+            resp = requests.post(self.api_url)
+
+        image_dims = {"width": "0", "height": "0"}
```
FrancescoSaverioZuppichini, line 77: why we are sending back this information (with `zeros`)

roboflow/models/instance_segmentation.py
```diff
@@ -3,13 +3,15 @@
 
 
 class InstanceSegmentationModel(InferenceModel):
-    def __init__(self, api_key, version_id):
+    def __init__(self, api_key, version_id, colors={}, preprocessing={}):
```
FrancescoSaverioZuppichini, line 6: Same as before

roboflow/models/object_detection.py
```diff
@@ -142,11 +150,27 @@ def predict(
...
+                        self.preprocessing["resize"]["width"]
+                    ) or dimensions[1] > int(self.preprocessing["resize"]["height"]):
+                        image = image.resize(
+                            (
+                                int(self.preprocessing["resize"]["width"]),
+                                int(self.preprocessing["resize"]["height"]),
+                            )
+                        )
+                        dimensions = image.size
+                        resize = True
```
FrancescoSaverioZuppichini, line 172: For the final version, could be either have more descriptive variables name for the condition or a little comment?

roboflow/models/object_detection.py
```diff
@@ -181,21 +209,213 @@ def predict(
...
+        Infers detections based on webcam feed from specified model
+
+        :param webcam_id: Webcam ID (default 0)
+        :param inference_engine_url: Inference engine address to use (default https://detect.roboflow.com)
+        :param in_jupyter: Whether or not to display the webcam within Jupyter notebook (default True)
+        :param confidence: Confidence threshold for detections
+        :param overlap: Overlap threshold for detections
+        :param stroke: Stroke width for bounding box
+        :param labels: Whether to show labels on bounding box
+        """
```
FrancescoSaverioZuppichini, line 288: Why are we using `sphinx` docstring style? Is it legacy?

roboflow/models/object_detection.py
```diff
@@ -181,21 +209,213 @@ def predict(
...
+        self.__generate_url(
+            confidence=confidence,
+            overlap=overlap,
+            stroke=stroke,
+            labels=labels,
+            inference_engine_url=inference_engine_url,
+        )
+
+        def plot_one_box(
+            x, img, color=None, label=None, line_thickness=None, colors={}
```
FrancescoSaverioZuppichini, line 302: same thing for `colors` here, as before

roboflow/models/object_detection.py
```diff
@@ -181,21 +209,213 @@ def predict(
...
+
+            if label in colors.keys():
+                color = colors[label]
+                color = color.lstrip("#")
+                color = tuple(int(color[i : i + 2], 16) for i in (0, 2, 4))
+
+            tl = (
+                line_thickness or round(0.002 * (img.shape[0] + img.shape[1]) / 2) + 1
+            )  # line/font thickness
+            # color = color or [random.randint(0, 255) for _ in range(3)]
```
FrancescoSaverioZuppichini, line 316: leftover :)

roboflow/models/object_detection.py
```diff
@@ -181,21 +209,213 @@ def predict(
...
+            for pred in predictions:
+                formatted_pred = [
+                    pred["x"],
+                    pred["y"],
+                    pred["x"],
+                    pred["y"],
+                    pred["confidence"],
+                ]
+
+                # convert to top-left x/y from center
```
FrancescoSaverioZuppichini, line 387: nice comment

roboflow/models/object_detection.py
```diff
@@ -181,21 +209,213 @@ def predict(
...
+            else:
+                cv2.imshow("Roboflow Webcam Inference", frame)
+                if cv2.waitKey(1) & 0xFF == ord("q"):  # quit when 'q' is pressed
+                    cap.release()
+                    break
+
+        cap.release()
+        if not within_jupyter:
+            cv2.destroyWindow("Roboflow Webcam Inference")
+            cv2.destroyAllWindows()
```
FrancescoSaverioZuppichini, line 417: nice

roboflow/util/general.py
```diff
@@ -0,0 +1,7 @@
+import sys
+
+
+def write_line(line):
+    sys.stdout.write("\r" + line)
+    sys.stdout.write("\n")
+    sys.stdout.flush()
```
FrancescoSaverioZuppichini, line 7: why are we doing this?

FrancescoSaverioZuppichini: why this file is named `general`?

roboflow/util/prediction.py
```diff
@@ -42,7 +42,7 @@ def plot_image(image_path):
     return figure, axes
 
 
-def plot_annotation(axes, prediction=None, stroke=1, transparency=60):
+def plot_annotation(axes, prediction=None, stroke=1, transparency=60, colors={}):
```
FrancescoSaverioZuppichini, line 45: same as before for `colors`

roboflow/util/prediction.py
```diff
@@ -53,7 +53,13 @@ def plot_annotation(axes, prediction=None, stroke=1, transparency=60):
...
     # Object Detection annotation
+
+    stroke_color = "r"
+
     if prediction["prediction_type"] == OBJECT_DETECTION_MODEL:
+        if prediction["class"] in colors.keys():
+            stroke_color = colors[prediction["class"]]
+
         # Get height, width, and center coordinates of prediction
         if prediction is not None:
```
FrancescoSaverioZuppichini, line 64: what will happen if `prediction` is `None`? Should it fail?

roboflow/util/prediction.py
```diff
@@ -78,9 +84,12 @@ def plot_annotation(axes, prediction=None, stroke=1, transparency=60):
...
+        if prediction["class"] in colors.keys():
+            stroke_color = colors[prediction["class"]]
+
         points = [[p["x"], p["y"]] for p in prediction["points"]]
         polygon = patches.Polygon(
-            points, linewidth=stroke, edgecolor="r", facecolor="none"
+            points, linewidth=stroke, edgecolor=stroke_color, facecolor="none"
         )
         axes.add_patch(polygon)
     elif prediction["prediction_type"] == SEMANTIC_SEGMENTATION_MODEL:
```
FrancescoSaverioZuppichini, line 95: No `else` in the end, this conditional flow seems to be incorrect

roboflow/util/prediction.py
```diff
@@ -93,7 +102,11 @@ def plot_annotation(axes, prediction=None, stroke=1, transparency=60):
 
 class Prediction:
     def __init__(
-        self, json_prediction, image_path, prediction_type=OBJECT_DETECTION_MODEL
+        self,
+        json_prediction,
+        image_path,
+        prediction_type=OBJECT_DETECTION_MODEL,
+        colors={},
```
FrancescoSaverioZuppichini, line 109: same as before

roboflow/util/prediction.py
```diff
@@ -179,6 +197,8 @@ def save(self, output_path="predictions.jpg", stroke=2, transparency=60):
             )
 
         elif self["prediction_type"] == CLASSIFICATION_MODEL:
+            if self["top"] in self.colors.keys():
```
FrancescoSaverioZuppichini, line 200: what's top?

roboflow/util/prediction.py
```diff
@@ -473,7 +497,9 @@ def json(self):
         return prediction_group_json
 
     @staticmethod
-    def create_prediction_group(json_response, image_path, prediction_type, image_dims):
+    def create_prediction_group(
```
FrancescoSaverioZuppichini, line 500: why this method is `static`? The correct way for factory methods is `classmethods`

roboflow/util/versions.py
```diff
@@ -45,7 +45,7 @@ def print_warn_for_wrong_dependencies_versions(
     dependencies_versions: List[Tuple[str, str, str]]
 ):
     wrong_dependencies_versions = get_wrong_dependencies_versions(dependencies_versions)
-    for (dependency, order, version, module_version) in wrong_dependencies_versions:
+    for dependency, order, version, module_version in wrong_dependencies_versions:
```
FrancescoSaverioZuppichini, line 48: why this change?

FrancescoSaverioZuppichini (commented):
I know this is a **draft**, I've added an extensive list of comments. Be sure to:

- never used immutable variables as default arguments
- when something has a big conditional flow, either name the conditional flow variable or add a comment for us

Thanks :rocket:

roboflow/core/version.py
Jacobsolawetz: In this case I think it is a dict

roboflow/models/classification.py
Jacobsolawetz: Ok, please level me up @FrancescoSaverioZuppichini, why is that?

roboflow/models/classification.py
Jacobsolawetz: this was probably to fix a bug - and dims usually not relevant with classification

I'll dig in

Jacobsolawetz:
* Better plotting display
* webcam stop button
* there is a likely bug with certain class colors in COCO

roboflow/config.py
Jacobsolawetz: It may indeed be written, but I like the cascade - going to keep it with the new comments

roboflow/util/prediction.py
Jacobsolawetz: "top" is the predicted classification class based on our schema there

roboflow/util/prediction.py
Jacobsolawetz: I think this prediction.py could use a large refactor in general, I will not attempt to defend it in this PR

roboflow/util/versions.py
Jacobsolawetz: Linter I think

requirements.txt
```diff
@@ -18,3 +18,4 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 wget
 requests_toolbelt
+ipython>=8.9.0
```
yeldarby, line 21: How heavy is this additional requirement? 

We probably should put some effort into really slimming down our list of requirements so installing our pip package isn’t such a burden (OpenCV and matplotlib in particular are beasts).

FrancescoSaverioZuppichini: we can have additional installs in `setup.py`, e.g. for `notebook` is setup. Once loaded, the package can ask the runtime where he is and download additional stuff 

I think opencv is needed in general, not sure why we have `matplotlib` is we have opencv/pillow

Jacobsolawetz:
@Jacobsolawetz test

roboflow/models/classification.py
FrancescoSaverioZuppichini: Copy, thanks. Resolving

roboflow/__init__.py
```diff
@@ -67,14 +69,117 @@ def auth(api_key):
...
+    if workspace is None:
+        write_line(
+            "visit " + APP_URL + "/auth-cli" " to get your authentication token."
+        )
+    else:
+        write_line(
+            "visit "
+            + APP_URL
+            + "/auth-cli/?workspace="
+            + workspace
```
yeldarby, line 93: Brad TODO: hook into this in core repo

roboflow/models/classification.py
yeldarby: Unresolving because it tripped me up in the notebook that this was here & had wrong values (also they're strings not numbers?)

roboflow/models/object_detection.py
```diff
@@ -142,11 +153,28 @@ def predict(
...
                 image = Image.open(image_path).convert("RGB")
                 dimensions = image.size
+                original_dimensions = copy.deepcopy(dimensions)
+
+                # Here we resize the image to the preprocessing settings before sending it over the wire
+                if "resize" in self.preprocessing.keys():
+                    if dimensions[0] > int(
+                        self.preprocessing["resize"]["width"]
+                    ) or dimensions[1] > int(self.preprocessing["resize"]["height"]):
+                        image = image.resize(
```
yeldarby, line 169: Looks like this only does `Stretch To` but should follow the same resize strat they preproc'd with.

roboflow/models/object_detection.py
```diff
@@ -181,21 +213,252 @@ def predict(
...
                 image_path=image_path,
                 prediction_type=OBJECT_DETECTION_MODEL,
                 image_dims=image_dims,
+                colors=self.colors,
             )
         # Returns base64 encoded Data
         elif self.format == "image":
             return resp.content
 
+    def webcam(
```
yeldarby, line 270: This is cool -- do we have regular video infer built into pip as well?

roboflow/__init__.py
```diff
@@ -118,6 +120,8 @@ def login(workspace=None, force=False):
 
     else:
         raise RuntimeError("Error logging in")
```
FrancescoSaverioZuppichini, line 122: This is not very description to user nor providing information if he opens a PR with the error, just call

```
r_login.raise_for_status()
```

and remove the condition

roboflow/__init__.py
```diff
@@ -118,6 +120,8 @@ def login(workspace=None, force=False):
...
+    
+    #return Roboflow().workspace()
```
FrancescoSaverioZuppichini, line 124: leftover

FrancescoSaverioZuppichini (changes requested):
One minor comment

roboflow/__init__.py
Jacobsolawetz: @yeldarby what this? I just authed via new account in incognito and it seemed pretty good

yeldarby: If you have multiple workspaces, selecting the one you need

Jacobsolawetz: Ok - I think it's OK for now, new user flow will be main concern for this method

roboflow/models/object_detection.py
Jacobsolawetz: We do not

roboflow/models/object_detection.py
Jacobsolawetz: ^^ @paulguerrie when we get a chance let's add img_dims back to classification response in infer server, that way we can know for hosted urls - @Tyler-Odenthal fixed for local classifications

paulguerrie (approved):
Looks great!  Pumped about this in the quick start notebooks!

PR #116, opened 2023-02-17
Classification annotation upload

ogjaylowe:
# Description

Updated annotation upload logic to support passing a class name as a raw string to classification projects.

## Type of change

test cases:
- empty string
- string is filepath that doesn't exist
- string is alphabet characters

For some reason using a number in a string such as `"1"` fails on upload and not sure why. Will need separate resolution.

-   [x] Docs updated? What were the changes: need to update docs to support classification string


Jacobsolawetz (approved):
LGTM! So you pass the class into the `annotation_path` string? Would be curious to see the docs when the come out of example usage

PR #117, opened 2023-02-22
small fix / better error message encountered during API debugging ses…

hansent:
couple of small fixes and better error logging that we came up during the API debugging session

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -289,7 +289,7 @@ class Project:
             # Construct URL for local image upload
             self.image_upload_url = "".join(
                 [
-                    "https://api.roboflow.com/dataset/",
+                    API_URL + "/dataset/",
                     project_name,
                     "/upload",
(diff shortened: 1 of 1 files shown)


roboflow/core/project.py
```diff
@@ -347,10 +347,13 @@ def __annotation_upload(self, annotation_path, image_id):
...
         annotation_string = open(annotation_path, "r").read()
         # Set annotation upload url
+
+        project_name = self.id.rsplit("/")[1]   
+
         self.annotation_upload_url = "".join(
             [
                 API_URL + "/dataset/",
-                self.name,
+                project_name,
```
hansent, line 356: this is what what causing the `Image not found in dataset.`. we saw sometimes. after having created a bunch of projects with the same name but different id/urls

Jacobsolawetz (approved):
🙏 🙏

PR #120, opened 2023-02-24
update annotation upload to reference correct project name

ogjaylowe:
# Description

Referenced `self.name` instead of `self.__project_name`. During original testing these coincidently were the same but can sometimes differ when we do string sanitation.

## Type of change

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -373,7 +373,7 @@ class Project:
         self.annotation_upload_url = "".join(
             [
                 API_URL + "/dataset/",
-                self.name,
+                self.__project_name,
                 "/annotate/",
                 image_id,
                 "?api_key=",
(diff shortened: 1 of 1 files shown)


FrancescoSaverioZuppichini:
Hi @ogjaylowe , thanks for the fix. Can you describe the issue better? Usually (even if there is badly used), `__` notation means "this is very internal". What `self.name` means and what `self.__project_name` means? 

Without a clear understanding of this, I cannot accept the changes. Sorry for the waiting, and thanks again for the PR

hansent:
ah, I actually ran into the same error while working on something else (but committed it to Jacobs Quickstart branch as it was part of that.

Definitly a bug in that it should use the project_url instead of the project_name

@ogjaylowe Do you know if self.__project_name is always the project_url?  The fix for it on the other branch used the parameter being passed in (split by `/`) similar to how its done in the upload function:  https://github.com/roboflow/roboflow-python/pull/109/commits/9cdf23cca2ae9bff1d7c62e09aed303f3009f354

hansent:
I checked and `self.__project_name` is the same as `id.rsplit("/")[1]`

hansent (approved):

paulguerrie (approved):
Seems legit

PR #123, opened 2023-03-03
YOLO Model Deploys

paulguerrie:
# Description

Updated deploy method of version class to support more model types.

## Type of change

Tested using local dev scripts.

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -312,16 +312,18 @@ class Version:
             model_path (str): File path to model weights to be uploaded
         """
 
-        supported_models = ["yolov8", "yolov5", "yolov7-seg"]
+        supported_models = ["yolov5", "yolov7", "yolov8"]
 
-        if model_type not in supported_models:
+        if not any(
+            supported_model in model_type for supported_model in supported_models
+        ):
             raise (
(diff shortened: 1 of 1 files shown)


roboflow/core/version.py
```diff
@@ -312,16 +312,18 @@ def deploy(self, model_type: str, model_path: str) -> None:
             model_path (str): File path to model weights to be uploaded
         """
 
-        supported_models = ["yolov8", "yolov5", "yolov7-seg"]
+        supported_models = ["yolov5", "yolov7", "yolov8"]
```
Jacobsolawetz, line 315: Is this up to date with what we ended up deciding to support?

Jacobsolawetz: On convert server side that is?

Jacobsolawetz (approved):
LGTM! Left some comments about supported types

roboflow/core/version.py
paulguerrie: I updated the list and changed yolov7 to yolov7-seg.  This will make sure yolov7 and yolov7-cls are rejected.  This logic may need some rework though since yolov5-cls would not be rejected but is also not supported.

PR #126, opened 2023-03-10
Classification Inference Fix - PredictionGroup

Tyler-Odenthal:
# Description

Had to add image_dims to the return statement for classification inference. The PredictionGroup requires that parameter and I would rather not delete it as it might be a legacy / active para in the PredictionGroup function.

Easiest fix was to add image.size to the classification predict function and pass that in the return to maintain the next function.

## Type of change

Tested on multiple two stage scripts. Confirmed that it does fix the classification inference issue.

## Docs

There is no change to the docs necessary and the call method will stay the same.

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -8,7 +8,7 @@ from roboflow.config import API_URL, APP_URL, DEMO_KEYS
 from roboflow.core.project import Project
 from roboflow.core.workspace import Workspace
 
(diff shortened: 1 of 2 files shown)


Jacobsolawetz (approved):
Looks great - good fix to get out and not wait for other branch to eventually come out - Cc @paulguerrie we should probably get img_dims into infer server response in case of other areas relying on it

PR #127, opened 2023-03-10
Removes TestPyPi from Publish Action

Jacobsolawetz:
# Description

Removes TestPyPi from Publish Action as we often do that manually ahead of time


Tyler-Odenthal (approved):
Sounds good, we can always add more test cases to the Publish Action later.

PR #129, opened 2023-03-27
Update config.py

alexvecchiettini:
Fixes "TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'"

# Description

Fixes "TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'"

I believe this Error appears in Windows when the path contains spaces.

## Type of change

I haven't tested it if not on my local system (Windows 11), it fixed the problem for me so I thought of submitting it to you who surely have a better understanding of the whole system and the eventual larger impacts this might have.

Hope it's useful!

### DIFF
--- a/roboflow/config.py
+++ b/roboflow/config.py
@@ -17,7 +17,7 @@ def get_conditional_configuration_variable(key, default):
     # default configuration location
     conf_location = os.getenv(
         "ROBOFLOW_CONFIG_DIR",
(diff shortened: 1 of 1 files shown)


Jacobsolawetz:
@alexvecchiettini thanks a million for finding - I'm going to add on a bit to store into `home_dir = os.getenv("HOME") or os.getenv("USERPROFILE")` in case Windows users want to `.login()` and keep that

PR #130, opened 2023-03-27
Readme refresh

Jacobsolawetz:
# Description

This PR introduces a new upper level `download_dataset` method and cleans up the README to be more up to date with what is possible and removes some of the older fluff

## Type of change

Docs update

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ print(dataset.location)
 
 ### Models
 
-Predict with any of over 50,000 public computer vision models. Train your own computer vision models on app.roboflow.com or train upload them from our - see https://github.com/roboflow/notebooks
+Predict with any of over 50,000 public computer vision models. Train your own computer vision models on app.roboflow.com or train upload your model from open source models - see https://github.com/roboflow/notebooks
 
 ```python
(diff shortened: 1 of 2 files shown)


mo-traor3-ai (commented):
roboflow.login() works, however, all other functions for accessing and downloading data and models does not work if you have not re-authed with the CLI since the last time you revoked your API key(s)

must use roboflow.login(force=True) to make the Python SDK functions work if you have revoked any API keys since your last CLI auth

this is not a problem/no bugs triggered if you had revoked API keys from workspaces you didn't auth with the CLI, prior to using the roboflow.login() method
image

SkalskiP:
As for @mo-traor3-ai comment about the `roboflow.login()`. I mentioned that on Slack, but it would be cool if we would not print the token. It is especially problematic in a notebook environment. 

As for the PR, it looks cool. I see some inspiration ;) One thing that is a bit off is the top banner. I use dark mode, and the bottom text is hardly visible.

Screenshot 2023-03-29 at 13 22 37

SkalskiP (approved):

PR #137, opened 2023-04-12
PT Weights Download

Jacobsolawetz:
# Description

API wrapping for the download of pytorch model weights

## Type of change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x ] New feature (non-breaking change which adds functionality)

PR #138, opened 2023-04-14
Pt download2

tonylampada:
# Description

This moves and renames a single method, as a follow-up of #137 

`Version.get_pt_weights` is replaced by `model.download()`, which makes more sense: download a model's weights is an operation that should belong to the model (version already has a `download()` method to download images)

Tested locally against the production api


paulguerrie (approved):
Looks great!

PR #139, opened 2023-04-18
add `is_prediction` parameter to upload functions

hansent:
The annotation upload API has a new parameter to mark whether the uploaded annotation data is a prediction rather than ground truth.

When annotation data is uploaded as a prediction, the image is kept in the batch or existing job it is in rather than being added directly to the dataset as an annotated image.  Using this parameter user can add annotation data but keep the image in the regular annotation flow for further review/adjustment of the annotations (e.g. for active learning workflows).

Could use some help testing this, I just put the code/parameters where I thought they needed to go for now.

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -346,7 +346,7 @@ class Project:
 
         return response
 
-    def __annotation_upload(self, annotation_path: str, image_id: str):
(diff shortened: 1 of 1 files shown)


Jacobsolawetz (approved):
Looks great!

PR #145, opened 2023-05-03
fix typo in config.py

hansent:
I think the config for the `OBJECT_DETECTION_URL` is reading the wrong env variable


Jacobsolawetz (approved):
Looks like it!

PR #146, opened 2023-05-03
use OBJECT_DETECTION_URL config value rather than hardcoded prod url for inference

hansent:
quick follow up to #145 to make it actually use the config value set by the environment variable rather than a hardcoded string


Jacobsolawetz (approved):
Yes!

Jacobsolawetz:
`make style`?

PR #148, opened 2023-05-17
`local` Parameter For Instance Segmentation Version

paulguerrie:
# Description

Added the `local` parameter functionality to instance segmentation giving the ability to infer using a local inference server instead of the hosted endpoint.

## Type of change

Tested locally with dev script

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -12,7 +12,7 @@ from roboflow.core.project import Project
 from roboflow.core.workspace import Workspace
 from roboflow.util.general import write_line
 
-__version__ = "1.0.8"
+__version__ = "1.0.9"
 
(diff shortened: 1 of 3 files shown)


Jacobsolawetz (approved):
Nice! Before publishing, remember to bump version

The tests should catch this, but would you also run a quick download of one of our export snippets on this code. I am having some dejavu of breaking that with adding a local var like this

paulguerrie:
> Nice! Before publishing, remember to bump version
> 
> The tests should catch this, but would you also run a quick download of one of our export snippets on this code? I am having some dejavu of breaking that with adding a local var like this

Nice Catch! Version has been bumped and I ran an export snippet without issue.

PR #150, opened 2023-05-17
Fix for issue #128 and #149

### LINKED ISSUE #128: Bug Report
roboflow/core/project.py, line 482, 

self.check_valid_image(image) should be self.check_valid_image(path)

image is only the filename.

### LINKED ISSUE #149: Failing "test_download_returns_dataset" unit test
In Windows 10 environment, when I ran the unit tests using `python -m unittest`, "test_download_returns_dataset" failed as shown in the screenshot below-
test_fail_roboflow-python

As we can see above, the the test fails at line 50 because the "my-spot" directory is saved as "D:\\my-spot" in Windows instead of "\my-spot" as normally done in Linux environment.
I think the code should check the existence of "my-spot" directory based on the OS as Windows uses "\\" for path seperation.

arijitde92:
# Description

Changed the input to self.check_valid_image() in line 489 of [project.py](https://github.com/roboflow/roboflow-python/blob/main/roboflow/core/project.py) from image to path as the function is meant to take image path but it was taking image variable as input. This fixes #128.

But after making the changes when I ran `python -m unittest`, "test_download_returns_dataset" was failing as the assertEqual() function was checking Linux style paths and not Windows style paths. Hence I created #149.

But I fixed #149 by changing the second arguments of line 46 and 50 of [test_version.py](https://github.com/roboflow/roboflow-python/blob/main/tests/test_version.py) such that the download path is asserted according to the OS environment. I used os.path.abspath as I saw that the download function in [version.py](https://github.com/roboflow/roboflow-python/blob/605fd458cd03ecaef6f46aa58c36150f74b56f83/roboflow/core/version.py#L203) is also using abspath for download location. Now all test cases are passing in the Windows environment.

## Type of change

Tested locally with dev script

No

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -486,7 +486,7 @@ class Project:
             images = os.listdir(image_path)
             for image in images:
                 path = image_path + "/" + image
-                if self.check_valid_image(image):
+                if self.check_valid_image(path):
                     self.single_upload(
                         image_path=path,
                         annotation_path=annotation_path,
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -43,11 +43,11 @@ class TestDownload(unittest.TestCase):
     def test_download_returns_dataset(self, *_):
         responses.add(responses.GET, self.api_url, json={"export": { "link": None }})
         responses.add(responses.GET, self.generating_url, json={"version": {"generating": False, "progress": 1.0}})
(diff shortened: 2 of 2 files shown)


arijitde92:
Hi @Jacobsolawetz, @FrancescoSaverioZuppichini, @Palashio, could you review the PR and let me know if any changes are required?

PR #176, opened 2023-08-31
Update config path resolution

### LINKED ISSUE #136: import roboflow
New bug: cannot import roboflow. From the error message seems the bug is at the get_conditional_configuration_variable method

image

capjamesg:
This PR uses the `os.path.join()` method for path resolution instead of manual concatenation of paths for generating the full path of the Roboflow config file. This will fix #136.


paulguerrie (approved):
LGTM

PR #179, opened 2023-09-01
Remove `wget` dependency and fix `JSONDecodeError` import

capjamesg:
# Description

This PR removes the `wget` dependency from `roboflow-python` and replaces its use with `requests`. This change affects downloading model weights and datasets from Roboflow. Both the dataset download and weight export should be tested to validate this change.

Also `requests.exceptions.JSONDecodeError` is deprecated and has been replaced with `json.JSONDecodeError`.

## Type of change

`pytest` was run to validate that all tests pass.

N/A

## Docs

N/A

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,8 +14,6 @@ requests
 six
 supervision
 urllib3>=1.26.6
-wget
 tqdm>=4.41.0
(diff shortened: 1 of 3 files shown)


paulguerrie (approved):
LGTM!

PR #182, opened 2023-09-07
fix: 🐛 collab deps conflict fix

onuralpszr:
### Description

Google collab has more recent version of packages and current versioning causing issue when you install it. This should install higher. 

List any dependencies that are required for this change.

## Type of change

Current Result

image

Fixed Result

image


onuralpszr:
@capjamesg  after merge a quick minor release should fix the problem as well.

capjamesg:
@paulguerrie Are there any concerns associated with removing the pinned version numbers and using the `>=` syntax?

paulguerrie:
> @paulguerrie Are there any concerns associated with removing the pinned version numbers and using the `>=` syntax?

It might be worth adding a `<` for the next major revision. For example `dependancy>=4.5.1,<5.0.0` to avoid breaking changes, but not totally necessary.

onuralpszr:
> > @paulguerrie Are there any concerns associated with removing the pinned version numbers and using the `>=` syntax?
> 
> It might be worth adding a `<` for the next major revision. For example `dependancy>=4.5.1,<5.0.0` to avoid breaking changes, but not totally necessary. 

Or we merge this in short term and I switch project  new toml for more controllable and add test for certain cases (just need login infos as secret so can do actually tests and  we can actually know in cases of breaking change until then we keep update as well

capjamesg:
Thank you for your contribution! Could you submit a PR that uses a `pyproject.toml` file and then we can revisit this issue? It would be great if the `pyproject.toml` file don't cause a user to upgrade past a new major version as per @paulguerrie's comment earlier. I am going to close this PR for now and we can follow up should you be interested in submitting a `pyproject.toml` PR.

PR #184, opened 2023-09-07
docs: 📝 adding supervision into readme section

onuralpszr:
## Docs

-   [X] Docs updated? What were the changes: Simple readme update for supervision

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
   <br>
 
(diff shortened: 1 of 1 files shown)


capjamesg (approved):
Nice quick change! I'm not sure why we didn't add `supervision` in here when we made this change 🤦

PR #190, opened 2023-09-20
Requests timeout fix

momoroboflow:
# Description

Implements timeout and max_retries for any functions that POST. Found out that there was an occasional hang due to network issues and the POST request would wait forever.

## Type of change

-   [X] This change requires a documentation update

Ran with a hosted model and confirmed that pipeline no longer gets stuck and does not fail for occasional network hiccups.

None, default arguments maintain original functionality

## Docs

-   [ ] Docs updated? What were the changes: Updated in-code docstrings; need to update GitBook

### DIFF
--- a/roboflow/models/classification.py
+++ b/roboflow/models/classification.py
@@ -55,13 +55,15 @@ class ClassificationModel:
         self.colors = {} if colors is None else colors
         self.preprocessing = {} if preprocessing is None else preprocessing
 
-    def predict(self, image_path, hosted=False):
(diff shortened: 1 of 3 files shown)


robiscoding:
We should allow the caller to handle how they want to retry failed calls to `predict`. Retry backoff can be helpful in some cases, but not where, for example, there is no internet connection. Additionally, this implementation duplicates the `__failsafe_post` method across multiple files, which is undesirable because it'll require changing them all anytime we change one.

PR #191, opened 2023-09-22
python version to >=3.8 from 3.6 in setup.py and readme

ryanjball:
# Description

Updated python version requirements to >=3.8 from 3.6 in setup.py and Readme. Python 3.8 is now the earliest version Roboflow supports because it’s the earliest version supported by our dependencies. Here’s [the discussion](https://github.com/roboflow/supervision/pull/180) from when supervision dropped support for 3.7 a few months ago.

## Type of change

-   [X] This change requires a documentation update

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


tonylampada (approved):
lgtm

PR #200, opened 2023-10-27
Feature: Video Inference

capjamesg:
# Description

This PR adds support for video inference to the Roboflow Python package.

## Type of change

_Will add a description once the draft PR is complete._

N/A

## Docs

We will publish:

1. A page on docs.roboflow.com that describes how video inference works.
2. A blog post announcing video inference with an example.
3. Pricing information.

### DIFF
--- a/roboflow/models/video.py
+++ b/roboflow/models/video.py
@@ -82,7 +82,7 @@ class VideoInferenceModel(InferenceModel):
 
             >>> model = project.version("1").model
 
-            >>> prediction = model.predict("video.mp4", fps=5, inference_type="object-detection"
+            >>> prediction = model.predict("video.mp4", fps=5, inference_type="object-detection")
         """
 
         url = urljoin(API_URL, "/video_upload_signed_url/?api_key=", self.__api_key)
--- a/roboflow/models/video.py
+++ b/roboflow/models/video.py
@@ -5,12 +5,9 @@ from urllib.parse import urljoin
 
 import magic
 import requests
-import os
 
 from roboflow.config import API_URL
 from roboflow.models.inference import InferenceModel
-
-MAXIMUM_VIDEO_SIZE = 1024 * 1024 * 1024 * 5 # 5GB
 
 SUPPORTED_ROBOFLOW_MODELS = ["object-detection", "classification", "instance-segmentation"]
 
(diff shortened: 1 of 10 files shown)


roboflow/models/video.py
```diff
@@ -0,0 +1,228 @@
...
+            >>> rf = roboflow.Roboflow(api_key="")
+
+            >>> project = rf.workspace().project("PROJECT_ID")
+
+            >>> model = project.version("1").model
+
+            >>> prediction = model.predict("video.mp4", fps=5, inference_type="object-detection"
+        """
+
+        url = urljoin(API_URL, "/video_upload_signed_url/?api_key=", self.__api_key)
```
PacificDou, line 88: I think this API will only return a signed_url for uploading, and user still need to upload the video by himself, right ? @bigbitbus

roboflow/models/video.py
```diff
@@ -0,0 +1,228 @@
...
+
+        url = urljoin(
+            API_URL, "/videoinfer/?api_key=", self.__api_key, "&job_id=", self.job_id
+        )
+
+        response = requests.get(url, headers={"Content-Type": "application/json"})
+
+        data = response.json()
+
+        if data["success"] != 0 or data["status_info"] != "success":
```
PacificDou, line 178: if int(data['status']) == 0:
    success
elif int(data['status']) == 1:
    still running
else:
    failed....will retry it if int(data['retries']) <= 3

roboflow/models/video.py
```diff
@@ -0,0 +1,228 @@
...
+        while True:
+            response = self.poll_for_response()
+
+            time.sleep(60)
+
+            print(f"({attempts * 60}s): Checking for inference results")
+
+            attempts += 1
+
+            if response != {}:
```
PacificDou, line 227: I'd suggest moving this check before sleep

roboflow/models/video.py
```diff
@@ -0,0 +1,228 @@
...
+
+            >>> model = project.version("1").model
+
+            >>> prediction = model.predict("video.mp4", fps=5, inference_type="object-detection"
+        """
+
+        url = urljoin(API_URL, "/video_upload_signed_url/?api_key=", self.__api_key)
+
+        if fps > 5:
+            raise Exception("FPS must be less than or equal to 5.")
```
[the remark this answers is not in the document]
bigbitbus: The user needs to know that the FPS can be upto the video frame rate, but given the fps-based pricing, set this value to as low as makes sense to their app. Maybe make it part of the blogpost?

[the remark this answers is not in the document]
capjamesg: It would be great if the server raised an exception if FPS is higher. Otherwise, if the user provides a video URL we have to download it in the package to check its FPS before we can do anything.

bigbitbus (approved):

PR #215, opened 2024-01-03
Add new commands to the CLI

tonylampada:
# Description

It's not ideally efficient for our users to need to install [roboflow-python](https://pypi.org/project/roboflow/) AND [roboflow-cli](https://www.npmjs.com/package/roboflow-cli) to do similar things.
It's also not ideal for us to need to maintain both ;-)

This PR adds more subcommands to the python version of the `roboflow` command:

* list workspaces and projects
* get details for workspaces and projects
* run inference

The commands are documented in [this new markdown file](https://github.com/roboflow/roboflow-python/blob/newcommands/CLI-COMMANDS.md) (and there's a reference to it in the main README)

It also fixes a bug with the authentication mechanism: if the user had authenticated for multiple workspaces, it would only use the API_KEY for the default workspace, and the authentication would fail for other workspaces. This fix benefits both the command line use case and the python-library use case.

Also, I was annoyed by the 88 line-length enforced by flake8. Increased it to 120. Some changes in this PR is only reformatting because of that :P

## Type of change

Installed locally and ran a bunch of commands

## Docs

* Added new markdown file in this repo: https://github.com/roboflow/roboflow-python/blob/newcommands/CLI-COMMANDS.md
* Eventually I will replace the [documentation for the node-cli](https://docs.roboflow.com/api-reference/install-cli) in our official docs, but I'd rather wait a bit for a few early adopters and make sure this doesn't have any obvious bugs.

### DIFF
--- a/CLI-COMMANDS.md
+++ b/CLI-COMMANDS.md
@@ -1,0 +1,23 @@
+## Available commands
+
+```bash
+roboflow --help
+usage: roboflow [-h] {login,download,upload,import,infer,project,workspace} ...
+
+Welcome to the roboflow CLI: computer vision at your fingertips 🪄
+
+options:
+  -h, --help            show this help message and exit
+
+subcommands:
+  {login,download,upload,import,infer,project,workspace}
+    login               Log in to Roboflow
+    download            Download a dataset version from your workspace or Roboflow Universe.
+    upload              Upload a single image to a dataset
+    import              Import a dataset from a local folder
+    infer               perform inference on an image
+    project             project related commands. type 'roboflow project' to see detailed command help
+    workspace           workspace related commands. type 'roboflow workspace' to see detailed command help
+```
+
+## Authentication
--- a/CLI-COMMANDS.md
+++ b/CLI-COMMANDS.md
@@ -1,7 +1,10 @@
-## Available commands
+## See available commands
 
 ```bash
-roboflow --help
+$ roboflow --help
+```
+
+```
 usage: roboflow [-h] {login,download,upload,import,infer,project,workspace} ...
 
 Welcome to the roboflow CLI: computer vision at your fingertips 🪄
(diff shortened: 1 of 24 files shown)


roboflow/roboflowpy.py
```diff
@@ -192,13 +255,103 @@ def _add_import_parser(subparsers):
...
+    workspaceget_parser.set_defaults(func=get_workspace)
+
+
+def _add_infer_parser(subparsers):
+    infer_parser = subparsers.add_parser(
+        "infer",
+        help="perform inference on an image",
+    )
+    infer_parser.add_argument(
+        "file",
```
bigbitbus, line 312: Why not a switch like `-f` like most of the other commands?

roboflow/roboflowpy.py
```diff
@@ -66,15 +71,83 @@ def upload_image(args):
...
 
 
+def list_projects(args):
+    rf = roboflow.Roboflow()
+    workspace = rf.workspace(args.workspace)
+    projects = workspace.project_list
+    for p in projects:
+        print()
+        print(p["name"])
+        print(f"  link: {APP_URL}/{p['id']}")
```
bigbitbus, line 81: Did you test for empty (newly created) projects with no versions, images etc.

bigbitbus (commented):
Looks really good. Had a couple of questions; happy to iterate after that.

roboflow/roboflowpy.py
tonylampada: Good call. Did that just now. Working as expected!

roboflow/config.py
```diff
@@ -81,21 +75,14 @@ def get_conditional_configuration_variable(key, default):
...
+    default_workspace = workspaces_by_url.get(default_workspace_url, None)
+    workspace = workspaces_by_url.get(workspace_url, default_workspace)
+    workspace = workspace or get_conditional_configuration_variable("RF_WORKSPACE", default=None)
+    if workspace:
+        return workspace.get("apiKey", None)
     if os.getenv("RF_API_KEY") is not None:
-        RF_API_KEY = os.getenv("RF_API_KEY")
-
-    return RF_API_KEY
+        return os["RF_API_KEY"]
```
hansent, line 88: look sliek the behavior changed here by pulling the returns into the ifs.  before, then env var would always override.  D you think thats behavior would want to maintain?  not sure if there are other things using the python package relying on it.

tonylampada: Ah, you're right. That wasn't my intention. I'll change the order here. Thanks Thomas!

roboflow/models/classification.py
```diff
@@ -143,10 +143,11 @@ def __generate_url(self):
         # Generates URL based on all parameters
         splitted = self.id.rsplit("/")
         without_workspace = splitted[1]
+        version = self.version or splitted[2]
```
hansent, line 146: do we enedd to check lenght here.  Cant see in this function what `self.id`.  looking at line 132 seems liek its setting `self.version` explicitly based on arg passed into `load_model`. If we're changing it, should we change it to take these via arguments instead?

CLI-COMMANDS.md
hansent, line 1: so nice! 💯

roboflow/models/classification.py
tonylampada: Good catch. Added the length check in there. Now version can come both in the id or the version parameter. If it comes in both, the version parameter overrides it.

tonylampada:
Thanks @bigbitbus and @hansent I'll send this to pypi and merge later!

PR #226, opened 2024-01-30
Allow downloads for non object-detection models

sberan:
# Description

Moves download logic into the base class to allow download of any model. Whether the model can be downloaded is controlled server-side, so in the future, model downloads can be added server-side without the need for client changes.

## Type of change

Tested in staging.

Will need to release a new version of the sdk.

## Docs

-   n/a

### DIFF
--- a/roboflow/models/inference.py
+++ b/roboflow/models/inference.py
@@ -5,6 +5,7 @@ import time
 import urllib
 from typing import List
 from urllib.parse import urljoin
+from tqdm import tqdm
 
 import requests
 from PIL import Image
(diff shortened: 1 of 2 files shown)


roboflow/models/inference.py
```diff
@@ -358,3 +359,43 @@ def poll_until_video_results(self, job_id) -> dict:
...
+
+        workspace, project, version = self.id.rsplit("/")
+
+        # get pt url
+        pt_api_url = f"{API_URL}/{workspace}/{project}/{self.version}/ptFile"
+
+        r = requests.get(pt_api_url, params={"api_key": self.__api_key})
+
+        r.raise_for_status()
+
```
bigbitbus, line 384: Why is the bar_progress removed?

PR #228, opened 2024-02-01
ordered-uploading

tonylampada:
# Description

The upload image endpoint has input parameters `sequence_order` and `sequence_size` that allows the client to control the parameter used to sort images in the backend.

This PR: 
* Exposes that as new parameters in `project.single_upload` method
* Automatically uses that in `workspace.upload_dataset` so that files are saved in the alphanumerical order they are found in the filesystem (instead of a random order).

So now if you run 
`$ roboflow import path/to/dataset/` 

you get something like this:

CleanShot 2024-02-01 at 15 50 47@2x

## Type of change

Tested by uploading it to staging from roboflow-python dev environment

Will release version 1.1.18

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -14,7 +14,7 @@ from roboflow.core.workspace import Workspace
 from roboflow.models import CLIPModel, GazeModel  # noqa: F401
 from roboflow.util.general import write_line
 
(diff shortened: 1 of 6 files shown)


hansent (approved):
lgtm, reviewed together on zoom

PR #234, opened 2024-02-27
Clarified `annotations_path`, `split params` in `project.upload`.

LinasKo:
# Description

This seems like a miniscule change, but it it would have saved me 2 hours today.

It took time to create a COCO annotations file, only to find that:
* RF API fails silently when you send a file with an incorrect format, without feedback as to what's wrong
* The parameter allows sending of the class label directly.

List any dependencies that are required for this change.

None

This only changes the docstring.

N/A

~~I got `mkdocs serve` running, but could not locate any `Project` docstring renderings.
If that is displayed somewhere, it should be checked that parameters of  `Project` are displayed correctly.~~

I think it's fine.

image

## Docs

-   [x] Docs updated? What were the changes:

Clarified a special case of `annotations_path`, explained it more thoroughly, and clarified the explanation for `split`.

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -371,10 +371,11 @@ class Project:
 
         Args:
             image_path (str): path to image you'd like to upload
-            annotation_path (str): if you're upload annotation, path to it
+            annotation_path (str): path to the annotation file. If not provided, the image will be uploaded without annotation.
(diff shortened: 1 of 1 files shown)


tonylampada (approved):
Thank you @LinasKo !

PR #237, opened 2024-03-05
Expand upload capabilities from the command line

tonylampada:
# Description

This PR expands the ability to upload datasets with different formats to roboflow.

* `-f FORMAT` parameter was removed. This tool is now capable of finding images and annotations regardless of format (new formats supported: coco, yolo* and createml)
* The trick for that is NOT about knowing every detail about each format (that's a job for Annotation.js). The trick is knowing just enough to understand what annotation data goes with each image file. Pair them together and let the backend figure it out.
* new parameter `-n batch_name` allows the user to specify batch name (which is also used as the jobName for annotated images)
* update `CLI-COMMANDS.md` with documentation for dataset import
* removes dependency from supervision (was required to convert datasets before uploading, but we're not doing that anymore)

## Type of change

- [x] Code maintenance (simplification of upload functionality)

Tested locally uploading a bunch of different things to production

## Docs

- updated `CLI-COMMANDS.md`
- Let's replace all references to roboflow-cli with this after this goes live!


roboflow/adapters/rfapi.py
```diff
@@ -103,6 +102,7 @@ def save_annotation(
     annotation_name: str,
     annotation_string: str,
     image_id: str,
+    job_name: str = DEFAULT_JOB_NAME,
```
hansent, line 105: should we add this to cmd arg also like batch name?

roboflow/core/workspace.py
```diff
@@ -273,9 +271,9 @@ def upload_dataset(
         dataset_path: str,
         project_name: str,
         num_workers: int = 10,
-        dataset_format: str = "yolov8",
```
hansent, line 276: this is always automatically detected now?

One thif about this vhnage is that it changes the function signature or user facing function.  Should we keep it and print a deprecation warning instead if parsed?

Not a problem if people are using named parameters, but if they have a script that passes the raw values in order, this might break existing scripts?

tests/datasets/sharks-tiny-createml/test/sharks_mp4-1_jpg.rf.4cfe68e1e1f580f36628ee5868b61564.jpg
hansent, line 1: do you know if these binary files get bundled as part of the pip package?  wondering about package size

tests/util/test_folderparser.py
hansent, line 1: ❤️ 💯

roboflow/config.py
```diff
@@ -71,13 +71,14 @@ def get_conditional_configuration_variable(key, default):
...
 DEFAULT_BATCH_NAME = "Pip Package Upload"
+DEFAULT_JOB_NAME = "Annotated via API"
 
 RF_WORKSPACES = get_conditional_configuration_variable("workspaces", default={})
 
 
 def load_roboflow_api_key(workspace_url=None):
     if os.getenv("RF_API_KEY") is not None:
-        return os["RF_API_KEY"]
+        return os.getenv("RF_API_KEY")
```
yeldarby, line 81: Is this a convention we use anywhere else? I think we call this `ROBOFLOW_API_KEY`in most spots... though I think in some spots we look in `.env` also and in some we look for a `.roboflow_key` file in the directory tree.

roboflow/adapters/rfapi.py
tonylampada: I don't think so. 
IMO the user either wants to set the name in this box or not.
The distinction between batch and job is a complexity the user shouldn't care at this point.

roboflow/core/workspace.py
tonylampada: Great point. Lemme fix that.

roboflow/config.py
tonylampada: yeah I'll change it to `ROBOFLOW_API_KEY` much better. 
`RF_API_KEY` isn't even working anyway.

tests/datasets/sharks-tiny-createml/test/sharks_mp4-1_jpg.rf.4cfe68e1e1f580f36628ee5868b61564.jpg
tonylampada: They don't.

* our `setup.py` excludes them
* I confirmed by installing roboflow in a fresh docker container =)
CleanShot 2024-03-07 at 10 38 40@2x

hansent (approved):
lgtm.  tested together on tuple using dev container

PR #238, opened 2024-03-08
Feature/yolonas upload

probicheaux:
# Description

Allows upload of yolonas to the platform! Note -- uploaders will need to create their opt.yaml file directly for yolonas since the deci repo doesn't create it, and it doesn't persist any of the data we need anywhere else. The cli should informo users how to do this, however. The model file is already weight_only=True compatible.

## Type of change

Staging! Deployed

Will need to deploy roboflow-model-conversion along with this.

## Docs

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -430,9 +430,9 @@ class Version:
 
         if not any(supported_model in model_type for supported_model in supported_models):
             raise (ValueError(f"Model type {model_type} not supported. Supported models are" f" {supported_models}"))
-        
+
         if "yolonas" in model_type:
-            self.deploy_yolonas(model_path)
+            self.deploy_yolonas(model_type, model_path)
             return
 
         if "yolov8" in model_type:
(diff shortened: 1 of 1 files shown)


roboflow/core/version.py
```diff
@@ -535,6 +540,74 @@ def deploy(self, model_type: str, model_path: str) -> None:
...
+                "The torch python package is required to deploy yolonas models."
+                " Please install it with `pip install torch`"
+            )
+
+        model = torch.load(os.path.join(model_path, "weights/best.pt"), map_location="cpu")
+        class_names = model["processing_params"]["class_names"]
+
+        opt_path = os.path.join(model_path, "opt.yaml")
+        if not os.path.exists(opt_path):
+            raise RuntimeError(
```
SolomonLake, line 559: this seems like it could be difficult for users to do.. I think many users run this .deploy from inside a colab notebook, do you have a sense for the UX of that? maybe we could add a section to our official notebook that creates the opt.yaml?

SolomonLake (approved):
one comment about usability, otherwise looks good!

roboflow/core/version.py
probicheaux: why would it be difficult for users to do this if they're in a colab? You can write to files with python

but yeah we can add this stuff to the notebook!

PR #248, opened 2024-04-05
BugFix: importing a yolo dataset with a trailing slash ignores annotations

tonylampada:
# Description

dumb bug:

```bash
roboflow import -w <workspaceId> -p <projectId> /path/to/some/yolo/format/dataset/
# uploads only images
```

```bash
roboflow import -w <workspaceId> -p <projectId> /path/to/some/yolo/format/dataset
# uploads images and annotations
```

This fixes it.

## Type of change

* New test case
* Tested uploading locally


probicheaux (approved):
great!

PR #252, opened 2024-04-26
Bugfix with .json not returning valid json and added .dict

Qfc9:
# Description
We are currently facing an issue with the `.predict` function which includes an option for `.json`. However, this `.json` method does not generate valid JSON data; rather, it returns a dictionary. This discrepancy is likely to cause confusion and issues for users.

The necessary adjustments are to be implemented in the `PredictionGroup` class within the utils module.

No additional dependencies are required for this change.

## Type of Change
- [X] Bug fix (non-breaking change which resolves an issue)
- [X] New feature (non-breaking change which introduces functionality)

## Testing and Validation
This change has been executed and verified through direct code implementation. The modification entails relocating the currently effective code to a `.dict()` method and subsequently converting its output into valid JSON within the `.json` method.

## Deployment Considerations
Potential issues might arise for users who are currently utilizing the output from the `.json` method (which is effectively a dictionary) as a dictionary in their code. These users may need to adjust their implementations to accommodate the changes.

### DIFF
--- a/roboflow/util/prediction.py
+++ b/roboflow/util/prediction.py
@@ -467,14 +467,17 @@ class PredictionGroup:
                     + ")"
                 )
 
-    def json(self):
-        prediction_group_json = {"predictions": []}
+    def json(self) -> str:
+        return json.dumps(self.dict())
+
+    def dict(self):
+        prediction_group_dict = {"predictions": []}
         for prediction in self.predictions:
-            prediction_group_json["predictions"].append(prediction.json())
+            prediction_group_dict["predictions"].append(prediction.json())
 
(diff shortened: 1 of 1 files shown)


capjamesg:
@Qfc9 Hello there! Do you have a code example showing a test case for this project?

iurisilvio:
I agree it might be misleading, but I think it is a common pattern in Python to call it `.json()` instead of `.dict()`. It is something like "data prepared to be JSON".

This change would be very backward incompatible with little value, I prefer to not do it, maybe adding type hints and better docs to the method would fix the confusion.

PR #260, opened 2024-06-20
Bump `opencv-python-headless` yanked version

iurisilvio:
# Description

We try to install a yanked version. Some installers fail because of that.

https://pypi.org/project/opencv-python-headless/#history

## Type of change

CI pass


capjamesg:
I have updated this PR to fix an issue where certifi and cycler were forcing the user to reload Colab after installing the package.

Issue #253, opened 2024-05-10
In Project#upload the argument batch_name has no effect


AlainPilon:
The following code:

```
project.upload(image_path='https://imgs.michaels.com/MAM/assets/1/726D45CA1C364650A39CD1B336F03305/img/A3A0E13DF0934FFDA4D0A3B01E2E99DE/004957.jpg', hosted_image=True, batch_name='will not work', tag_names=['one', 'two'])
```

will not create a new batch under the Annotation section. It will just put the uploaded file under the "Uploaded from API" batch.

Also the file will not be tagged.

joaomarcoscrs:
Seems like the `_hosted_upload_url` method is forgetting to consider both `batch_name` and `tag_names` when building the URL to upload the images. It should work properly for local hosted images though.

https://github.com/roboflow/roboflow-python/blob/013512b0054d1c25e5e40af548d64193afec3f08/roboflow/adapters/rfapi.py#L160

I created this pull request, think that should solve it:

https://github.com/roboflow/roboflow-python/pull/268

iurisilvio:
Fixed by #268! 🎉

PR #268, opened 2024-06-25
Fixes batch_name and tag_names not being set when uploading a hosted image

joaomarcoscrs:
# Description

Adds support to set `batch_name` and `tag_names` when uploading a hosted image, it only work for local images right now. Also adds some tests for the `upload_image` method in `rfapi.py` adapter

From this issue: https://github.com/roboflow/roboflow-python/issues/253

## Type of change

Added unit tests for the `rfapi.py` adapter, testing the `upload_image` method output for multiple scenarios.

## Docs

-   [x] Docs didn't have to be updated


roboflow/adapters/rfapi.py
```diff
@@ -158,22 +158,37 @@ def _save_annotation_url(api_key, project_url, name, image_id, job_name, is_pred
...
     url = f"{API_URL}/dataset/{project_url}/upload?api_key={api_key}"
-    url += f"&name={os.path.basename(image_path)}&split={split}"
-    url += f"&image={urllib.parse.quote_plus(image_path)}"
+
+    for key, value in kwargs.items():
+        if isinstance(value, list):
+            url += "".join([f"&{str(key)}={str(v)}" for v in value])
+        else:
+            url += f"&{str(key)}={str(value)}"
+
```
iurisilvio, line 169: I think `urllib.parse.urlencode` does exactly it.

```
>>> from urllib.parse import urlencode
>>> urlencode({"a": 1, "b": ["x", "y", "z"]}, doseq=True)
'a=1&b=x&b=y&b=z'
```

```suggestion
    if kwargs:
        querystring = urlencode(kwargs, doseq=True)
        url += f"&{querystring}"
```

roboflow/adapters/rfapi.py
```diff
@@ -158,22 +158,37 @@ def _save_annotation_url(api_key, project_url, name, image_id, job_name, is_pred
...
+        split=split,
+        image=urllib.parse.quote_plus(image_path),
+        batch_name=batch_name,
+        tag=tag_names,
+    )
+
+
 def _local_upload_url(api_key, project_url, batch_name, tag_names, sequence_number, sequence_size, kwargs):
-    url = f"{API_URL}/dataset/{project_url}/upload?api_key={api_key}&batch={batch_name}"
+    query_params = dict(batch_name=batch_name, tag=tag_names, **kwargs)
```
iurisilvio, line 186: The URL used a `batch` param instead of `batch_name`. Is the param wrong or you confused names here? I think `batch` is correct, but didn't investigate the API.

```suggestion
    query_params = dict(batch=batch_name, tag=tag_names, **kwargs)
```

roboflow/adapters/rfapi.py
joaomarcoscrs: Yes, it even quotes via `quote_plus`, eliminating the need to explicitly encode image paths using it. Changed at 21a26e21673285dee6884e6df3e8d5251ca0b79f

roboflow/adapters/rfapi.py
joaomarcoscrs: Great catch, fixed at 21a26e21673285dee6884e6df3e8d5251ca0b79f

iurisilvio (approved):

PR #276, opened 2024-06-27
version.model should return None if no model

stellasphere:
# Description

There should be no instance where `version.model` should return a model type class when there is no valid model present.

Fixing this issue: https://roboflow.slack.com/archives/C056ZUJPK08/p1719359598748219

## Type of change

Locally
and here: https://colab.research.google.com/drive/1rnYv49x4oVWt--O3alw2wg6pr2qNapEO?usp=sharing

## Docs


roboflow/core/version.py
```diff
@@ -101,6 +101,10 @@ def __init__(
 
             version_without_workspace = os.path.basename(str(version))
 
+            version_info = requests.get(
+                API_URL + "/" + workspace + "/" + project + "/" + self.version + "?api_key=" + self.__api_key
```
LinasKo, line 105: Nitpick: I'd use an f-string, but I don't consider it a blocker.

roboflow/core/version.py
```diff
@@ -136,6 +140,13 @@ def __init__(
             else:
                 self.model = None
 
+            # check if version has a model
+            if version_info.status_code == 200:
+                version_info = version_info.json()["version"]
+
+                if not bool(version_info["models"]):
```
LinasKo, line 147: `if` calls `bool` for you. You don't need to do it explicitly.

roboflow/core/version.py
```diff
@@ -136,6 +140,13 @@ def __init__(
...
+                    self.model = None
```
LinasKo, line 148: This implementation removes any prior `model` value if there's no model of this version. If that's correct, I suggest moving this up, approx. to line 105, and returning early. No reason to instantiate `model` object if it will be reset afterwards.

LinasKo (changes requested):
What I cannot evaluate is whether prior behavior, creating the `model` object without a matching model on the api, is valid use case somewhere.

roboflow/core/version.py
stellasphere: 🤦 yeah

stellasphere:
Thanks @LinasKo !

> What I cannot evaluate is whether prior behavior, creating the `model` object without a matching model on the api, is valid use case somewhere.

I strongly believe that creating a model object when there is no model present is invalid behavior because the object it creates, a child of an `InferenceModel`, and its methods `predict` and `load_model` will fail because it doesn't exist. Doing so also makes it impossible to determine from the package that a model does or does not exist.

LinasKo:
One other request: there's a `test.py`, which I think was included by accident.

stellasphere:
Yup, I need to revoke my key too. Thanks for catching!

LinasKo (approved):
Approving, with the caveat that I did not run it locally, nor ran any tests.

stellasphere:
Sorry, fixed a failing test. @LinasKo can you re-approve?

LinasKo:
M'afraid test.py snuck its way into the commit again :wink:

What I'm seeing is - the model may not exist, but that's valid only in a project version where it's not been trained yet. Is that right?

stellasphere:
> M'afraid test.py snuck its way into the commit again 😉
You would think i would learn, but no.

> the model may not exist, but that's valid only in a project version where it's not been trained yet. Is that right?
Yes, where it's not been trained. (or uploaded, but the same thing)

LinasKo (approved):

iurisilvio (approved):
I agree there is the risk someone use this behaviour, but I think it should be fixed, "no model" is a known behavior and we have to return `None` to make it explicit.

LGTM! 🎉

PR #279, opened 2024-07-02
Fix types in version init

iurisilvio:
# Description

Fix errors found after mypy merge.

```
roboflow/core/version.py:110: error: Unsupported operand types for in ("str" and "Response")  [operator]
roboflow/core/version.py:110: error: Value of type "Response" is not indexable  [index]
```

It was a small bug that would be triggered only when our API failed.

Minor regression from https://github.com/roboflow/roboflow-python/pull/276.

## Type of change

CI pass.

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -101,13 +101,12 @@ class Version:
 
             version_without_workspace = os.path.basename(str(version))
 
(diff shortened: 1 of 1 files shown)


stellasphere (approved):
Better implementation, LGTM!

PR #285, opened 2024-07-12
chore: Enhance type checking with mypy and improve code quality

cdragos:
This PR enhances our code quality checks and type safety:

1. Integrates mypy into our `check_code_quality` process
2. Improves path handling using `pathlib.Path` for better cross-platform support
3. Fixes several type-related issues caught by mypy
4. Updates `.dockerignore`

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -59,20 +59,13 @@ def check_key(api_key, model, notebook, num_retries=0):
         return "onboarding"
 
 
-def auth(api_key):
-    r = check_key(api_key)
-    w = r["workspace"]
-
-    return Roboflow(api_key, w)
-
-
 def login(workspace=None, force=False):
     os_name = os.name
 
     if os_name == "nt":
-        default_path = os.path.join(os.getenv("USERPROFILE"), "roboflow/config.json")
+        default_path = os.path.join(os.getenv("USERPROFILE", ""), "roboflow/config.json")
     else:
-        default_path = os.path.join(os.getenv("HOME"), ".config/roboflow/config.json")
+        default_path = os.path.join(os.getenv("HOME", ""), ".config/roboflow/config.json")
 
     # default configuration location
     conf_location = os.getenv(
@@ -143,7 +136,7 @@ def initialize_roboflow(the_workspace=None):
 
     conf_location = os.getenv(
         "ROBOFLOW_CONFIG_DIR",
-        default=os.getenv("HOME") + "/.config/roboflow/config.json",
+        default=os.getenv("HOME", "") + "/.config/roboflow/config.json",
     )
 
     if not os.path.isfile(conf_location):
@@ -176,7 +169,7 @@ def load_model(model_url):
         project = path_parts[2]
         version = int(path_parts[-1])
     else:
-        raise ("Model URL must be from either app.roboflow.com or universe.roboflow.com")
+        raise ValueError("Model URL must be from either app.roboflow.com or universe.roboflow.com")
 
     project = operate_workspace.project(project)
     version = project.version(version)
@@ -204,7 +197,7 @@ def download_dataset(dataset_url, model_format, location=None):
         version = int(path_parts[-1])
         the_workspace = path_parts[1]
     else:
-        raise ("Model URL must be from either app.roboflow.com or universe.roboflow.com")
+        raise ValueError("Model URL must be from either app.roboflow.com or universe.roboflow.com")
     operate_workspace = initialize_roboflow(the_workspace=the_workspace)
 
     project = operate_workspace.project(project)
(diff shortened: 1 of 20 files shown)


iurisilvio:
Why pyright instead of mypy? I don't have strong opinions on which one is better, but I don't like to change tools without good reasons.

I'm happy we now have type checking, but I don't think changing tools will improve it a lot. I prefer to keep iterating on mypy config and enable many options still disabled.

cdragos:
@iurisilvio Thanks for your feedback. I apologize for not providing a description earlier.

I agree about being cautious with changing tools. While exploring `pyright`, I found it caught several issues with its default settings that the current setup missed:

- Syntax errors like `raise ("Model URL must be from either app.roboflow.com or universe.roboflow.com")` 
- Runtime errors such as `r = check_key(api_key)` where the function signature doesn't match
- Incorrect return type annotations, e.g., `def train(..) -> bool:` where the actual return type differs

One feature I found particularly useful is pyright's Inferred Return Types, similar to TypeScript. Unlike `mypy`, which marks unannotated return types as `Any`, pyright provides the actual return type.

Many more differences [documented here](https://microsoft.github.io/pyright/#/mypy-comparison) made me think `pyright` is a better tool.

To move forward, I propose the following alternatives:

1. Run both tools in parallel to evaluate which better suits our project needs
2. Keep `mypy` and some `changes / fixes` from my PR

Let me know what you think!

roboflow/__init__.py
```diff
@@ -282,7 +275,7 @@ def project(self, project_name, the_workspace=None):
 
         dataset_info = dataset_info.json()["project"]
 
-        return Project(self.api_key, dataset_info)
+        return Project(self.api_key or "", dataset_info)
```
iurisilvio, line 278: I don't think this is the fix. `api_key` could be `None` for public projects.

I prefer to ignore the error instead of fix the type and cover a possible bug.

roboflow/models/inference.py
```diff
@@ -2,8 +2,8 @@
 import json
 import os
 import time
-import urllib
```
iurisilvio, line 5: I think pyright is enforcing this change, but it is mostly cosmetic, with minor performance gains. I prefer to keep it scoped on `urllib` for readability.

roboflow/models/video.py
```diff
@@ -65,7 +65,7 @@ def __init__(
         """  # noqa: E501 // docs
         self.__api_key = api_key
 
-    def predict(  # type: ignore[override]
+    def predict(
```
iurisilvio, line 68: This is an example where mypy do better. This override with incompatible signature is bad.

I don't know how to fix it without breaking compatibility, but I like to have it annotated as an issue.

roboflow/__init__.py
```diff
@@ -59,20 +59,13 @@ def check_key(api_key, model, notebook, num_retries=0):
         return "onboarding"
 
 
-def auth(api_key):
```
iurisilvio, line 62: It is backward incompatible. I'd prefer to keep this function.

iurisilvio (commented):
A few comments to ship this PR, considering we'll keep mypy.

roboflow/__init__.py
cdragos: Okay, but I am curious about the use case, at this moment the calling of this function raises an exception
`TypeError: auth() missing 1 required positional argument: 'api_key'`

cdragos:
@iurisilvio Thank you for your thorough feedback and guidance throughout this PR process. I've addressed all the comments raised. 🏄🏼

roboflow/__init__.py
```diff
@@ -59,20 +59,13 @@ def check_key(api_key, model, notebook, num_retries=0):
...
-    w = r["workspace"]
-
-    return Roboflow(api_key, w)
-
-
 def login(workspace=None, force=False):
     os_name = os.name
 
     if os_name == "nt":
-        default_path = os.path.join(os.getenv("USERPROFILE"), "roboflow/config.json")
```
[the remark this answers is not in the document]
cdragos: I used [Path.home](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home). Should handle the correct case, both for windows and linux.

roboflow/__init__.py
iurisilvio: Ah ok , I didn't realized it is totally broken. 👍🏻 Yes, we can remove it.

iurisilvio: Can you remove it again? 😅 I reviewed all the PR already (all good), I'll wait for this last change.

cdragos: Sure! Done!

iurisilvio (approved):

iurisilvio:
Thanks! 🚢

PR #292, opened 2024-07-19
Ban module level imports for `cv2`, `matplotlib` and `numpy`

iurisilvio:
# Description

Ban some big global imports to make it faster to `import roboflow`. It makes initial import ~3x faster.

## Type of change

[Benchmark](https://colab.research.google.com/drive/12bXEnQljCl47gDhQbJmSxNkS8gQp0v0C?usp=sharing)

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import copy
 import json
 import os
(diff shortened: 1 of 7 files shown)


roboflow/core/version.py
```diff
@@ -37,6 +37,11 @@
 from roboflow.util.general import write_line
 from roboflow.util.versions import get_wrong_dependencies_versions, print_warn_for_wrong_dependencies_versions
 
+if TYPE_CHECKING:
+    import numpy as np
+
+    from roboflow.models.inference import InferenceModel
+
 load_dotenv()
```
tonylampada, line 45: TIL about python's type checking and that it doesn't affect runtime.

tonylampada (approved):
lgtm

PR #304, opened 2024-08-07
chardet is not required for anything

tonylampada:
# Description

Removes chardet dependency.
* not used by our code
* doesn't show up as a dependency of anything in [pipdeptree](https://pypi.org/project/pipdeptree/)
* tests pass without it
* command line still seems to work
* git history does not provide a clear reason for why it was added

## Type of change

-   [x] Maintainability

* Tests passing
* Used the CLI locally

No need to release because of this

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 certifi
-chardet==4.0.0
 idna==3.7
 cycler
 kiwisolver>=1.3.1


tonylampada:
Yeah, response.text seems to work ok without it too
CleanShot 2024-08-07 at 11 05 35@2x

capjamesg (approved):

PR #312, opened 2024-08-20
Refactoring dataset upload

lrosemberg:
# Description

In PR #266, the way it was implemented changes the way the CLI displays the output, and also causes an error when trying to parse the response from upload_image.

In addition, some methods like `project.single_upload` had too many responsibilities, which made it difficult to handle exceptions correctly.

That said, what does this PR do?

- Shifts the responsibility for parsing the error correctly to the rfapi layer
- Makes only the rfapi layer throw exceptions
- Creates new methods `project.upload_image` and `project.save_annotation`, these start to be used inside `project.single_upload` and `workflow.upload_dataset`, isolating a good part of the code and keeping the functions with single responsibility
- Fixes the CLI output
- Fixes the bug introduced in @266 that fails to parse some image upload responses (more information here: https://github.com/roboflow/roboflow-python/pull/266#issuecomment-2295388263)
- Creates some test cases to validate `project.upload`

- Created some tests in the code to validate the `project.upload`
- Downloaded a dataset, edited a annotation to fail, and uploaded using the CLI, results below:

CleanShot 2024-08-19 at 23 50 31@2x

CleanShot 2024-08-20 at 00 07 22@2x


tonylampada (approved):
FANTASTIC.
Thank you @lrosemberg

PR #316, opened 2024-08-21
no print, only log

asymptotic:
# Description

No `print`, only `log`: I'm working on a project that uses `roboflow-python` and rather than making changes to suppress stdout for this module, I updated it to use a Python logger instead.

Every module gets its own logger using `logging.getLogger(__name__)`.

Everything is logged as INFO. 

The logger uses f-strings because the optimization benefit for not using them really doesn't matter here. Ruff now ignores G003 and G004 as a result.

## Type of change

Ruff and mypy checks pass.

I know there are still interactive prompts in there; for my next trick, I might figure out something to do with them for when running in a non-interactive environment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,8 @@ ignore = [
     "FIX",
     "FLY",
     "FURB",
+    "G003",
+    "G004",
     "ISC",
     "N",
     "PERF",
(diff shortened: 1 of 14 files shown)


asymptotic:
let me do some more updates and I'll resubmit on a different branch!

PR #323, opened 2024-09-02
Import paligemma format into text-image-pairs project

tonylampada:
# Description

If you export a text-image-pairs dataset in the paligemma format, and import it again, only the images will be imported, but not the annotations.

This PR makes the CLI able to read jsonl annotations and send them to the API backend.

## Type of change

* New test for local parser
* Tested on staging

* Release after corresponding change on the API (https://github.com/roboflow/roboflow/pull/4221 ✅)


roboflow/util/folderparser.py
```diff
@@ -197,12 +203,27 @@ def _loadAnnotations(folder, annotations):
...
 
+def _read_jsonl(path):
+    data = []
+    with open(path) as file:
+        for line in file:
+            try:
+                json_object = json.loads(line.strip())
+                data.append(json_object)
+            except json.JSONDecodeError:
+                print(f"Warning: Skipping invalid JSON line in {path}")
```
iurisilvio, line 223: Skip empty lines (last line for example), because it is not valid JSON and it is not a real issue.

Also, it is easy to include invalid line number.

```suggestion
        for linenum, line in enumerate(file, 1):
            if not line:
                continue
            try:
                json_object = json.loads(line.strip())
                data.append(json_object)
            except json.JSONDecodeError:
                print(f"Warning: Skipping invalid JSON line in {path}:{linenum}")
```

iurisilvio (approved):

iurisilvio:
You have to fix lint.

PR #330, opened 2024-10-02
Fix check for has_model for model upload

SolomonLake:
# Description

Fixes bug introduced here: https://github.com/roboflow/roboflow-python/pull/279/files, where `has_model` is always false for models uploaded to Roboflow. Those models will not have a "models" field on their version record.

## Type of change

Tested locally with uploaded model, and model trained with roboflow.

no

## Docs


tonylampada (approved):
lgtm!

PR #334, opened 2024-10-11
bugfix - downloading yolov8 files incorrectly amends data.yaml

tonylampada:
# Description

Our train notebooks like [train-yolov8-object-detection-on-custom-dataset.ipynb](https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/train-yolov8-object-detection-on-custom-dataset.ipynb) are having to do weird yaml amending like

CleanShot 2024-10-11 at 14 31 41

When you export a dataset in yolov8 format from the UI, you get a correct yaml that has

```yaml
train: ../train/images
val: ../valid/images
test: ../test/images
```

But if download it with roboflow-python you get a modified yaml like

```yaml
test: ../test/images
train: football-players-detection-12/train/images
val: football-players-detection-12/valid/images
```

It's not good that the train notebook needs to undo the mistake of roboflow-python

This PR fixes that mistake.

## Type of change

Tested locally

### DIFF
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -7,7 +7,6 @@ import shutil
 import sys
 import time
 import zipfile
-from importlib import import_module
 from typing import TYPE_CHECKING, Optional, Union
 
 import requests
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -233,7 +233,7 @@ class Version:
 
         self.__download_zip(link, location, model_format)
         self.__extract_zip(location, model_format)
-        self.__reformat_yaml(location, model_format)
+        self.__reformat_yaml(location, model_format)  # TODO: is roboflow-python a place to be munging yaml files?
 
         return Dataset(self.name, self.version, model_format, os.path.abspath(location))
 
@@ -944,7 +944,7 @@ class Version:
                 content["train"] = location + content["train"].lstrip(".")
                 content["val"] = location + content["val"].lstrip(".")
                 content["test"] = location + content["test"].lstrip(".")
-            if format in ["yolov5pytorch", "yolov7pytorch", "yolov8", "yolov9"]:
+            if format in ["yolov5pytorch", "yolov7pytorch"]:
                 content["train"] = location + content["train"].lstrip("..")
                 content["val"] = location + content["val"].lstrip("..")
             try:
--- a/roboflow/core/version.py
+++ b/roboflow/core/version.py
@@ -206,19 +206,6 @@ class Version:
 
         self.__wait_if_generating()
 
-        if model_format == "yolov8":
-            # if ultralytics is installed, we will assume users will want to use yolov8 and we check for the supported version  # noqa: E501 // docs
-            try:
-                import_module("ultralytics")
-                print_warn_for_wrong_dependencies_versions([("ultralytics", "==", "8.0.196")])
-            except ImportError:
-                print(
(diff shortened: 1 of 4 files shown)


roboflow/core/version.py
```diff
@@ -246,7 +232,7 @@ def download(self, model_format=None, location=None, overwrite: bool = False):
 
         self.__download_zip(link, location, model_format)
         self.__extract_zip(location, model_format)
-        self.__reformat_yaml(location, model_format)
+        self.__reformat_yaml(location, model_format)  # TODO: is roboflow-python a place to be munging yaml files?
```
LinasKo, line 235: I speculate, based on 2 days of experience with this:

In the past - yes. `data.yml` used to contain `path`, providing a path to the dataset. This would be based on where the dataset is located on the system. Now, yolov8 and yolo11 do not require that, but I don't know about other models.

roboflow/core/version.py
```diff
@@ -957,7 +943,7 @@ def data_yaml_callback(content: dict) -> dict:
                 content["train"] = location + content["train"].lstrip(".")
                 content["val"] = location + content["val"].lstrip(".")
                 content["test"] = location + content["test"].lstrip(".")
-            if format in ["yolov5pytorch", "yolov7pytorch", "yolov8", "yolov9"]:
```
LinasKo, line 960: We should test yolov9 as well. I thiiiiink it was yolov5-based? Could be that it needs special treatment like this.

roboflow/core/version.py
```diff
@@ -206,19 +205,6 @@ def download(self, model_format=None, location=None, overwrite: bool = False):
...
-                import_module("ultralytics")
-                print_warn_for_wrong_dependencies_versions([("ultralytics", "==", "8.0.196")])
-            except ImportError:
-                print(
-                    "[WARNING] we noticed you are downloading a `yolov8` datasets but you don't have `ultralytics` installed. "  # noqa: E501 // docs
-                    "Roboflow `.deploy` supports only models trained with `ultralytics==8.0.196`, to intall it `pip install ultralytics==8.0.196`."  # noqa: E501 // docs
-                )
-                # silently fail
-                pass
-
```
LinasKo, line 221: I'd prefer to remove rather than keep, but let me paint an alternative.

This serves 3 purposes:
* Tell user if the version is too new (e.g. known to not work, but only at this time point)
* Tell user if they're not using ultralytics that we assume they need it
* Tell the maintainer that ultralytics is required.

In the ideal world, we'd instead soften the wording, saying something akin to "When working with yolov8, roboflow works best with ultralytics<=8.2.103. You can install it with ..."

This way, the message is not exactly incorrect in the future, but still gives good info to users.

Leaving it to you to decide what we do. Removal is also fine, in my opinion 🙂

tonylampada: mmm idk, IMO this is the wrong place to do this. 
Theres many reasons a user might want to download a version.
It's too weird to be checking ultralytics version as part of downloading a zip file.

roboflow/core/version.py
tonylampada: This will not fix the path attribute. It will leave the train/test/val attributes untouched with the original content that was downloaded.
There's [this other PR](https://github.com/roboflow/Annotation.js/pull/119) on Annotation.js that will remove that.

roboflow/core/version.py
tonylampada: Yeah, let's release this and we can test that ;-)

LinasKo (approved):

LinasKo:
Than you very much @tonylampada!

There's an initiative in making the notebooks more reliable next week - we'll test it then.

PR #354, opened 2025-01-20
Added Confidence Argument to keypoint detection model

Greenstan:
# Description

A new "Confidence" argument has been added to the keypoint detection model (based on the confidence argument found in the object detection model)

## Type of change

A Dedicated keypoint detection model test file has been made with multiple test cases including one where the confidence is changed


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
SolomonLake: can you remove this file?

roboflow/models/keypoint_detection.py
```diff
@@ -150,6 +153,7 @@ def __generate_url(self):
                 self.base_url + without_workspace + "/" + str(version),
                 "?api_key=" + self.__api_key,
                 "&name=YOUR_IMAGE.jpg",
+                f"&confidence={self.confidence}",
```
SolomonLake: going to want `str(self.confidence)`

.github/dependabot.yml
Greenstan: This was an auto-generated file by the PR workflow of the project. I have manually deleted it now

roboflow/models/keypoint_detection.py
Greenstan: I've fixed this string concatenation method to be similar to the api_key string concat structure.

```python
[
...
"&confidence=" + str(self.confidence),
]
```

tests/models/test_keypoint_detection.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.version_id = f"{self.workspace}/{self.dataset_id}/{self.version}"
+
+    def test_init_sets_attributes(self):
+        instance = KeypointDetectionModel(self.api_key, self.version_id, version=self.version)
+
+        self.assertEqual(instance.id, self.version_id)
+        self.assertEqual(instance.api_key, self.api_key)
```
SolomonLake: KeypointDetectionModel doesn't have api_key on instance, and there are several other errors here -- can you run `python -m unittest` and make sure tests are passing?

roboflow/models/keypoint_detection.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         id: str,
         name: Optional[str] = None,
         version: Optional[str] = None,
+        confidence: Optional[int] = 10,
```
SolomonLake: let's default to `40`, that's the standard in this repo

tests/models/test_keypoint_detection.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual(len(result.predictions), 1)
+        self.assertEqual(result.predictions[0].confidence, 0.544)
+
+    @responses.activate
+    def test_predict_with_confidence(self):
+        instance = KeypointDetectionModel(self.api_key, self.version_id, version=self.version)
+
+        responses.add(responses.POST, self.api_url, json=MOCK_RESPONSE, status=200)
+
+        result = instance.predict("tests/images/MM2A_46_R_T.jpg", confidence=30)
```
SolomonLake: if you want to be able to pass confidence into .predict (which is a good idea), you will need to add it to predict arguments.

roboflow/models/keypoint_detection.py
Greenstan: The Default confidence has been set to 40 now.

tests/models/test_keypoint_detection.py
Greenstan: The confidence argument has been added to the predict method as a kwarg with a default of None (if no confidence argument is added to the .predict() method, it just uses the model instance confidence [self.confidence])

Greenstan: The confidence argument is then passed down to the _generate_url method similar to how it is done in other models in this project.

tests/models/test_keypoint_detection.py
Greenstan: I found some unnecessary assertEquals that didn't really need to be tested. I updated all the tests to make sure they are passing 

Now all tests pass on my end (apart from the test_predict_error_response test which is supposed to fail)

tests/models/test_keypoint_detection.py
```diff
@@ -0,0 +1,71 @@
...
+from dotenv import load_dotenv
+from requests.exceptions import HTTPError
+
+from roboflow.models.keypoint_detection import KeypointDetectionModel
+from roboflow.util.prediction import PredictionGroup
+
+load_dotenv(Path("../../.env"))
+
+
+with open(Path("../annotations/keypoint-detection-annotations/MM2A_46_R_T_predictions.json")) as f:
```
SolomonLake: tests get run from root -- can you update:

```suggestion
with open(Path("tests/annotations/keypoint-detection-annotations/MM2A_46_R_T_predictions.json")) as f:
```

tests/models/test_keypoint_detection.py
```diff
@@ -0,0 +1,71 @@
...
+    MOCK_RESPONSE = json.load(f)
+
+
+class TestKeypointDetection(unittest.TestCase):
+    api_key = os.getenv("ROBOFLOW_API_KEY")
```
SolomonLake:
```suggestion
    api_key = os.getenv("ROBOFLOW_API_KEY", "test-api-key")
```

tests/models/test_keypoint_detection.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertEqual(instance.version, self.version)
+        self.assertEqual(instance.base_url, "https://detect.roboflow.com/")
+
+    @responses.activate
+    def test_predict_local_image(self):
+        instance = KeypointDetectionModel(self.api_key, self.version_id, version=self.version)
+
+        responses.add(responses.POST, self.api_url, json=MOCK_RESPONSE, status=200)
+
+        result = instance.predict("/roboflow-python/tests/images/MM2A_46_R_T.png")
```
SolomonLake:
```suggestion
        result = instance.predict("tests/images/MM2A_46_R_T.png")
```

and change the other .png references too

tests/models/test_keypoint_detection.py
```diff
@@ -0,0 +1,71 @@
...
+        request = responses.calls[0].request
+        self.assertEqual(request.params["confidence"], "30")
+
+    @responses.activate
+    def test_predict_error_response(self):
+        instance = KeypointDetectionModel(self.api_key, self.version_id, version=self.version)
+
+        responses.add(responses.POST, self.api_url, json={"error": "Invalid API key"}, status=401)
+
+        with self.assertRaises(HTTPError):
```
SolomonLake:
```suggestion
        with self.assertRaises(Exception):
```

do this to avoid failing test

SolomonLake (changes requested):
can you make sure you are running the tests locally using `python3 -m unittest` (or `python`) from the root of repo?

You can see test failures here: https://github.com/roboflow/roboflow-python/actions/runs/16078304796/job/45497459911?pr=354

I left suggested fixes which are all going to be necessary.

Greenstan:
@SolomonLake The Local path issues have been updated according to your suggestions as well as the switch to "with self.assertRaises(Exception):" and the env api-key variable.

SolomonLake (approved):
Looks good to me, and tested it locally. Thanks for the contribution!

PR #355, opened 2025-01-27
Model processor

lrosemberg:
# Description

- Created `model_processor.py` which processes and zips models before uploading.
- Moved the processing responsibility from `version.py` to `model_processor.py`.
- Added the `-f` flag to the `login` command to force re-login, even if already logged in.

## Type of change

- [x] Maintance

- Locally pointing to staging

### DIFF
--- a/roboflow/roboflowpy.py
+++ b/roboflow/roboflowpy.py
@@ -16,10 +16,7 @@ from roboflow.models.semantic_segmentation import SemanticSegmentationModel
 
 
 def login(args):
-    if args.force:
-        roboflow.login(force=True)
-    else:
-        roboflow.login()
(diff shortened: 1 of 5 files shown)


roboflow/roboflowpy.py
```diff
@@ -16,7 +16,10 @@
 
 
 def login(args):
-    roboflow.login()
+    if args.force:
+        roboflow.login(force=True)
+    else:
+        roboflow.login()
```
iurisilvio, line 22: I guess you can trust argparse here.

```suggestion
    roboflow.login(force=args.force)
```

iurisilvio (dismissed):

tonylampada (approved):
lgtm

PR #363, opened 2025-03-10
Add a project method for creating annotation jobs

shantanubala:
# Description

This PR adds a method onto the `Project` class for creating an annotation job that is assigned to specific users.

## Type of change

Added unit tests and verified the HTTP request against the endpoint.

N/A

## Docs

We would need to update this page: https://docs.roboflow.com/api-reference/annotation-jobs

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -812,14 +812,9 @@ class Project:
         image_details = data["image"]
 
         return image_details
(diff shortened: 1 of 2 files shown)


SolomonLake (approved):
lgtm, definiton matches endpoint and test looks great!

PR #367, opened 2025-03-25
Fix handling of invalid annotation descriptions and add tests to dataset uploads

shantanubala:
# Description

This PR fixes a user-reported bug that occurs when uploading a dataset - sometimes, annotation data is an empty string for an image and causes uploads to fail with the following error:

> [ERR] grain-multi/datasets/train/images/0011.png ('str' object has no attribute 'get')

## Type of change

Added unit tests, also tested locally with a dataset

No

## Docs

N/A


lrosemberg (approved):
LGTM!

PR #368, opened 2025-03-25
Add methods for retrieving batches for a project

shntu:
# Description

I talked to a user today who was trying to upload images via API, but was wondering if there was a way to retrieve upload batches via the Python SDK. It turns out there's no Python SDK function for fetching the batches, so I've implemented them in this PR with unit tests.

## Type of change

Added unit tests and tested the SDK locally.

No

## Docs

I'll update these docs if the PR looks good: https://docs.roboflow.com/api-reference/batches


joaomarcoscrs (approved):
Code LGTM!

PR #382, opened 2025-05-27
Upgrade devcontainer

tonylampada:
# Description

Upgrades the development container environment by updating Python from 3.8 to 3.10 and integrating uv as the package manager for faster dependency installation and better development experience.

Necessary because vscode debugger doesn't like python3.8 anymore

## Type of change

- [x] Maintenance (devx)

I can debug stuff in the devcontainer again :-)

Nope

## Docs

N/A


lrosemberg (approved):
LGTM

PR #407, opened 2025-08-11
Fixing format export for CLI/python sdk

digaobarbosa:
# Description

The format that was used to export before calling the train did not follow the current logic that exists in the backend. And the backend did not use modelType to calculate which link to train on.

## Type of change

locally, staging

None

## Docs

-   [ ] Docs updated? What were the changes: None


tonylampada (approved):
LGTM

PR #412, opened 2025-09-09
Add are_predictions flag to dataset upload

sberan:
# Description

This flag allows dataset uploads to be sent as predictions (not automatically added to the dataset).

## Type of change

Unit tests cover this case

## Docs

-   will need to update [relevant documentation](https://docs.roboflow.com/developer/upload-a-dataset)

### DIFF
--- a/roboflow/core/workspace.py
+++ b/roboflow/core/workspace.py
@@ -287,7 +287,7 @@ class Workspace:
         project_type: str = "object-detection",
         batch_name=None,
         num_retries=0,
-        are_predictions=False,
+        is_prediction=False,
     ):
         """
         Upload a dataset to Roboflow.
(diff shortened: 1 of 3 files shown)


roboflow/core/workspace.py
```diff
@@ -287,6 +287,7 @@ def upload_dataset(
         project_type: str = "object-detection",
         batch_name=None,
         num_retries=0,
+        are_predictions=False,
```
iurisilvio: I'd prefer to be consistent and keep using `is_prediction`.

sberan: totally makes sense. I updated.

iurisilvio (approved):
LGTM!

You probably want to bump version before merge.

PR #463, opened 2026-04-24
Soft-delete / Trash support: SDK + CLI

yeldarby:
## Summary

Adds project, version, and workflow lifecycle management mirroring the soft-delete and Trash features added to the web app in [roboflow/roboflow#11131](https://github.com/roboflow/roboflow/pull/11131). Deleting a project, version, or workflow now moves it to Trash (30-day retention) and cancels any in-flight training jobs on it; items can be restored within the retention window.

- Project-level: `Project.delete()` / `Project.restore()`
- Version-level: `Version.delete()` / `Version.restore()`
- Workspace-level: `Workspace.trash()`, `Workspace.restore_from_trash()`
- CLI: `roboflow project delete|restore`, `roboflow version delete|restore`, `roboflow workflow delete|restore`, `roboflow trash list`

The SDK doesn't yet have a first-class `Workflow` handle object, so workflow lifecycle is exposed through `rfapi.delete_workflow()` + `Workspace.restore_from_trash("workflow", id)` plus the matching CLI commands.

## Details

**New rfapi functions** (`roboflow/adapters/rfapi.py`):
- `delete_project(api_key, workspace_url, project_url)` — `DELETE /:workspace/:project`
- `delete_version(api_key, workspace_url, project_url, version)` — `DELETE /:workspace/:project/:version`
- `delete_workflow(api_key, workspace_url, workflow_url)` — `DELETE /:workspace/workflows/:workflow`
- `list_trash(api_key, workspace_url)` — `GET /:workspace/trash`
- `restore_trash_item(api_key, workspace_url, item_type, item_id, parent_id=None)` — `POST /:workspace/trash/restore`

**SDK methods** — `Project.delete()` returns the server response. `Project.restore()` looks the project up by slug in the workspace Trash and restores it (raises `RuntimeError` if not found). Same shape for `Version`. For scripts without a live handle (or for workflows, which don't have an SDK object yet), `Workspace.trash()` returns the list of items and `Workspace.restore_from_trash(type, id, parent_id)` restores by id.

**CLI commands** — destructive commands (`project delete`, `version delete`, `workflow delete`) prompt for confirmation interactively and accept `--yes` / `-y` for scripted use. All commands honor `--json` for structured output and use `output_error()` with actionable hints on every error path (per the Agent Experience checklist in `CONTRIBUTING.md`).

**Error handling** — `rfapi` extracts the `error` field from JSON response bodies before raising `RoboflowError`, so users see a clean message ("Not authorized to view trash") rather than the raw response body.

[...]

PR #468, opened 2026-04-29
CLI: polish three rough edges from the 1.3.7 prod shake-down

yeldarby:
## Why

End-to-end testing of the v1.3.7 CLI against prod ([roboflow#11131](https://github.com/roboflow/roboflow/pull/11131) / [roboflow-python#463](https://github.com/roboflow/roboflow-python/pull/463) shipped this morning) surfaced three rough edges. None are correctness bugs — every command does the right server-side thing — but they all lead users (and especially agents) into wrong conclusions and were called out in the prod shake-down report. Worth fixing before the next release.

### 1. Auth errors return exit 3 (should be 2)

Per `CONTRIBUTING.md`: `0=success, 1=error, 2=auth, 3=not-found`. Every `rfapi.RoboflowError` from the trash endpoints was caught with `exit_code=3`, so scripts can't differentiate `re-auth` from `wrong slug`:

```bash
$ roboflow -k bogus-key trash list --json ; echo $?
{"error": ...}
3                ← should be 2
```

### 2. `--json` silently bypasses destructive-action confirmation

```bash
$ roboflow project delete moondream-example --json < /dev/null
{"deleted": true, "trash": true, ...}    ← deleted, no prompt!
```

Old logic was `if not yes and not json: prompt`. That conflated *output formatting* with *destructive intent* — anyone piping `… --json | jq` got their project nuked because `--json` was treated as "I'm in CI, skip the prompt."

### 3. Idempotent re-delete bubbles up a misleading "missing scope" 404

When a project/version/workflow is already in Trash, the public API's URL filter excludes it from the active set, so the next `DELETE` returns 404 ("Unsupported request. `DELETE /ws/proj` does not exist..."). The CLI surfaced that as exit 3 with `Check your API key has 'project:update' scope` — factually wrong (the user has the scope; the item is just where they wanted it).

## What

All three fixes live centrally so future trash / destructive endpoints inherit the right behavior for free.

**Auth exit code (`#1`)**
- `RoboflowError(message, status_code=None)` now stores `status_code`. Defaulted to `None` for back-compat with the 30+ existing call sites that pass message-only.
- `_raise_for_trash_response` stamps `response.status_code` on the raised exception.
- New `output_api_error(args, exc, hint=..., auth_hint=..., not_found_hint=...)` helper in `cli/_output.py` does the central status→exit-code mapping (401→2, 404→3, else→1). All four trash handlers (project / version / workflow / trash) replace hardcoded `output_error(..., exit_code=3)` with it.

**Destructive guard (`#2`)**
- New `confirm_destructive(args, prompt)` in `cli/_output.py`:
  - `--yes` → proceed.
  - TTY without `--yes` → prompt (regardless of `--json`).
  - No TTY without `--yes` → bail with exit 1 + hint about `--yes`.
- `project delete`, `version delete`, `workflow delete` all switched to it.

**Idempotent re-delete (`#3`)**
- Each delete handler intercepts a 404 from `rfapi.delete_X`, probes `list_trash`, and if the slug/url is found in the matching `sections` array, emits a synthetic success payload mirroring a real delete plus `alreadyInTrash: true`. If the slug is genuinely missing from trash too, the 404 propagates as exit 3 (typos still surface as errors).

Sample payload from a re-delete:

```json
{
  "deleted": true,
  "type": "project",
  "workspace": "q3-board-meeting",
  "project": "moondream-example",
  "projectId": "Jjk3Tjq2yjxenyA9g5er",
  "trash": true,
  "alreadyInTrash": true
}
```

## Tests

`tests/cli/test_trash_polish.py` adds 14 tests covering all three:
- `output_api_error` mapping for 401 / 404 / 500 / no-status-code (defaults to 1, NOT 3 or 2 — important so legacy `RoboflowError(text)` callers don't accidentally claim "not found" / "auth").
- `_raise_for_trash_response` actually attaches `status_code`.
- `confirm_destructive` honors `--yes`, refuses no-TTY without `--yes`, ignores `--json` when deciding (regression guard for finding #2), and prompts via typer on a TTY (both accept and decline paths).
- Idempotent re-delete success path on project / version / workflow, plus the negative case where a 404 with empty trash propagates as exit 3 (so we don't mask real "not found" errors).

```
$ python -m unittest discover -s tests
Ran 526 tests in 7.222s
OK (skipped=1)

$ ruff check roboflow tests
All checks passed!

$ ruff format --check roboflow tests
113 files already formatted

$ mypy roboflow
Success: no issues found in 62 source files
```

## Backwards compatibility

- `RoboflowError(message)` (no status_code) still works — every call site outside the trash endpoints does this and is unaffected.
- The new helpers (`output_api_error`, `confirm_destructive`) are additive in `_output.py`. Nothing else imports them yet.
- Wire format of successful deletes is unchanged; the new `alreadyInTrash` field is additive for the previously-error-out 404 case, so any caller currently parsing a successful delete payload keeps working.

## Test plan

- [x] `pytest` / `python -m unittest` — all 526 tests pass.
- [x] `ruff check roboflow tests` — clean.
- [x] `ruff format --check roboflow tests` — clean.
- [x] `mypy roboflow` — clean.
- [ ] One reviewer sanity-check of the synthetic `alreadyInTrash` payload shape (mirrors successful delete fields verbatim — `type`, `workspace`, `project`/`workflow`, `projectId`/`workflowId`, `trash`).

### DIFF
--- a/roboflow/adapters/rfapi.py
+++ b/roboflow/adapters/rfapi.py
@@ -12,7 +12,18 @@ from roboflow.config import API_URL, DEFAULT_BATCH_NAME, DEFAULT_JOB_NAME
 
 
 class RoboflowError(Exception):
-    pass
+    """Generic API error.
+
+    Optional `status_code` is the HTTP status from the upstream response
+    when available — set by helpers like `_raise_for_trash_response` so
+    callers can branch on auth (401) vs not-found (404) without string
+    matching the message. Existing call sites that pass only a message
+    still work; the attribute defaults to `None`.
+    """
+
+    def __init__(self, message, status_code=None):
+        super().__init__(message)
+        self.status_code = status_code
 
 
 class ImageUploadError(RoboflowError):
@@ -911,8 +922,11 @@ def _raise_for_trash_response(response):
     error messages are agent-friendly. Falls back to the raw text if the
     body isn't JSON or doesn't contain an `error` field.
 
-    The single `raise` at the end means we can't accidentally swallow the
-    intended error if a future refactor widens the except clause.
+    Carries the HTTP status code on the raised exception so callers (e.g.
+    the CLI) can map auth failures (401) to the right exit code without
+    string-matching. The single `raise` at the end means we can't
+    accidentally swallow the intended error if a future refactor widens
+    the except clause.
     """
     msg = None
     try:
@@ -922,7 +936,7 @@ def _raise_for_trash_response(response):
     except ValueError:
         # Body wasn't JSON — fall through to response.text.
         pass
-    raise RoboflowError(msg or response.text)
+    raise RoboflowError(msg or response.text, status_code=response.status_code)
 
 
 def delete_project(api_key, workspace_url, project_url):
--- a/roboflow/adapters/rfapi.py
+++ b/roboflow/adapters/rfapi.py
@@ -29,17 +29,15 @@ class RoboflowError(Exception):
 class ImageUploadError(RoboflowError):
     def __init__(self, message, status_code=None):
         self.message = message
-        self.status_code = status_code
         self.retries = 0
-        super().__init__(self.message)
+        super().__init__(self.message, status_code=status_code)
 
 
 class AnnotationSaveError(RoboflowError):
     def __init__(self, message, status_code=None):
         self.message = message
-        self.status_code = status_code
         self.retries = 0
-        super().__init__(self.message)
+        super().__init__(self.message, status_code=status_code)
 
 
 def get_workspace(api_key, workspace_url):
--- a/roboflow/adapters/devicesapi.py
+++ b/roboflow/adapters/devicesapi.py
@@ -38,8 +38,7 @@ class DeviceApiError(RoboflowError):
     """Raised when a device API call returns a non-success status."""
 
     def __init__(self, message: str, status_code: Optional[int] = None) -> None:
-        self.status_code = status_code
-        super().__init__(message)
+        super().__init__(message, status_code=status_code)
 
 
 class DeviceNotFoundError(DeviceApiError):
--- a/roboflow/cli/_output.py
+++ b/roboflow/cli/_output.py
@@ -176,6 +176,93 @@ def output_error(
     sys.exit(exit_code)
 
 
+def confirm_destructive(args: Any, prompt: str) -> bool:
+    """Gate a destructive action on either ``--yes`` or an interactive TTY confirmation.
+
+    Returns ``True`` if the action is approved (caller should proceed) or
+    ``False`` if the user declined at the prompt (caller should bail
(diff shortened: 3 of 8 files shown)


digaobarbosa (commented):
Quick suggestions to address the codex review detail.

roboflow/adapters/rfapi.py
digaobarbosa:
```suggestion
        self.message = message
        self.retries = 0
        super().__init__(self.message, status_code=status_code)
```

digaobarbosa:
```suggestion
        self.message = message
        self.retries = 0
        super().__init__(self.message, status_code=status_code)
```

probicheaux:
@claude review this. don't nitpick

roboflow/adapters/rfapi.py
```diff
@@ -12,7 +12,18 @@
...
+    Optional `status_code` is the HTTP status from the upstream response
+    when available — set by helpers like `_raise_for_trash_response` so
+    callers can branch on auth (401) vs not-found (404) without string
+    matching the message. Existing call sites that pass only a message
+    still work; the attribute defaults to `None`.
+    """
+
+    def __init__(self, message, status_code=None):
+        super().__init__(message)
+        self.status_code = status_code
```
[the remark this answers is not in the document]
probicheaux: Fixed in 7c7037e — `ImageUploadError.__init__` and `AnnotationSaveError.__init__` now forward `status_code` to `super().__init__(...)` so the parent no longer overwrites the subclass-provided HTTP code with `None`. Added a `TestRoboflowErrorSubclassStatusCode` regression class to `tests/cli/test_trash_polish.py` covering both subclasses (with and without an explicit `status_code`). Thanks for catching this.

roboflow/adapters/rfapi.py
probicheaux: Fixed in 7c7037e — `ImageUploadError.__init__` and `AnnotationSaveError.__init__` now forward `status_code` to `super().__init__(...)` so the parent no longer overwrites the subclass-provided HTTP code with `None`. Added a `TestRoboflowErrorSubclassStatusCode` regression class to `tests/cli/test_trash_polish.py` covering both subclasses (with and without an explicit `status_code`). Thanks for catching this.

probicheaux: Fixed in 7c7037e — `ImageUploadError.__init__` and `AnnotationSaveError.__init__` now forward `status_code` to `super().__init__(...)` so the parent no longer overwrites the subclass-provided HTTP code with `None`. Added a `TestRoboflowErrorSubclassStatusCode` regression class to `tests/cli/test_trash_polish.py` covering both subclasses (with and without an explicit `status_code`). Thanks for catching this.

digaobarbosa (dismissed):
LGTM

iurisilvio (approved):

PR #469, opened 2026-05-01
Add device management SDK and CLI

rvirani1:
# Description

Adds Python SDK + CLI bindings for the new external Deployments / Device Management API (added to the platform in roboflow/roboflow#11350 and hardened in #11441). Closes [ENT-1179](https://linear.app/roboflow/issue/ENT-1179/sdk-add-device-management-apis-light-full).

The SDK wraps every route documented in `docs/api/deployments/overview.md` on roboflow/roboflow master — 1:1, no invented surface. PATCH `/config`, stream commands, device delete, and fleet-groups CRUD are explicitly out of scope per the doc's "Versioning and scope" section.

**New surface:**

- `Workspace.devices()`, `Workspace.device(id)`, `Workspace.create_device(...)`
- `Device` class with `config()`, `config_history()`, `streams()`, `stream(id)`, `logs(...)`, `telemetry(...)`, `events(...)`, `refresh()`
- `roboflow device {list, get, create, config, config-history, streams, stream, logs, telemetry, events}` with `--json`, exit codes 0/1/2/3, rate-limit hints
- Typed exceptions in `roboflow/adapters/devicesapi.py`: `DeviceBadRequestError` (400), `DeviceAuthError` (401/403), `DeviceNotFoundError` (404), `DeviceRateLimitedError` (429), `DeviceApiError` (5xx)

**Files:**
- `roboflow/adapters/devicesapi.py` (new) — HTTP layer for all 10 routes
- `roboflow/core/device.py` (new) — `Device` model
- `roboflow/core/workspace.py` — wired in `devices()` / `device(id)` / `create_device(...)`
- `roboflow/cli/handlers/device.py` (new) + registration in `roboflow/cli/__init__.py`
- `tests/test_device.py` (new), `tests/cli/test_device_handler.py` (new)
- `CLI-COMMANDS.md` — quickstart entries

**Dependencies:** none new — uses `requests` and `typer` already required by the SDK.

## Type of change

- [x] This change requires a documentation update

**Unit tests** — 33 new tests pass (`python -m unittest tests.test_device tests.cli.test_device_handler`). Full repo suite: 545/545 pass. Coverage:
- URL building for every route; CSV serialization for `service`/`severity`; opaque cursor round-trip for events
- Error mapping for 400/401/403/404/404+GraphMethodException/429/500 → typed exceptions
- `Device` method dispatch; `Workspace.devices()`/`device(id)`/`create_device` integration
- CLI: subcommand registration, `--json` and text output, exit-code mapping (404→3, 401→2, 429→1)

**Lint / type:** `ruff check roboflow tests` clean, `ruff format --check roboflow` clean, `mypy roboflow` clean (no new issues).

**Live staging validation** against `api.roboflow.one` workspace `device-manager-demo-workspace`

| # | Endpoint | Result |
|---|---|---|
| 1 | `GET /:ws/devices/v2` | OK — 96 devices, schema matches `mappers.js` |
| 2 | `POST /:ws/devices/v2` | Not exercised (avoid mutating shared workspace); covered by unit tests |
| 3 | `GET /:ws/devices/v2/:id` | OK — schema match |
| 4 | `GET .../config` | OK — 16 keys incl. `environment_variables`, `services` |
| 5 | `GET .../config/history` | OK — pagination roundtrip works |
| 6 | `GET .../streams` | OK — schema match |
| 7 | `GET .../streams/:sid` | OK — schema match |
| 8 | `GET .../logs` | OK — schema match |
| 9 | `GET .../telemetry?time_period=1h` | OK — 31 buckets, schema match |
| 10 | `GET .../events` | Server-side 500 when no filter; works with `entity_type=stream`. Not an SDK bug — flagged separately. |

Negative cases against staging — every one mapped to the correct typed exception with the correct status code: bad `time_period` (400), bad events `cursor` (400), bad logs `start_time` (400), unknown deviceId (404), invalid api_key (401).

CLI parity verified end-to-end: `roboflow device list --json` → 96 entries; `roboflow device get <id>` → human-readable text; `roboflow device telemetry <id> --time-period 1h --json` → 31 buckets; 404 → exit 3; 401 → exit 2 with the `device:read` scope hint.

## Will the change affect Universe? If so was this change tested in universe?

No. Pure-Python SDK and CLI changes only; no Universe surface touched.

Nope

### Infrastructure impact

- [ ] This change affects infrastructure needs (GPUs, Cloud Function sizing, storage etc.)

## Docs

- [x] Docs updated? What were the changes:
  - `CLI-COMMANDS.md` — added device commands to the quickstart and the command-groups table
  - Inline docstrings on the `Device` class call out that `config()` is sensitive (may include `environment_variables` and integration credentials)
  - Follow-up doc PR for the full reference will land in `roboflow-product-docs` per the repo's CLI documentation policy

### DIFF
--- a/roboflow/adapters/devicesapi.py
+++ b/roboflow/adapters/devicesapi.py
@@ -20,6 +20,19 @@ from roboflow.config import API_URL
 
 DEFAULT_TIMEOUT = (10, 60)
 
+# Cap on raw error-body bytes surfaced through exception messages. Server-side
+# 500s sometimes return a multi-KB HTML stack trace; without a cap the whole
+# blob would land in `str(exc)` and wreck terminals/log lines.
+_MAX_ERROR_BODY_CHARS = 1024
+
+
+def _truncate(text: str) -> str:
+    if not text:
+        return text
+    if len(text) <= _MAX_ERROR_BODY_CHARS:
+        return text
+    return text[:_MAX_ERROR_BODY_CHARS] + "…[truncated]"
+
 
 class DeviceApiError(RoboflowError):
     """Raised when a device API call returns a non-success status."""
@@ -78,6 +91,7 @@ def _raise_for_status(response: requests.Response) -> None:
             message = response.text
     except Exception:  # noqa: BLE001
         message = response.text
+    message = _truncate(message)
     code = response.status_code
     if code == 400:
         raise DeviceBadRequestError(message or "Bad request", status_code=code)
--- a/roboflow/adapters/devicesapi.py
+++ b/roboflow/adapters/devicesapi.py
@@ -18,6 +18,8 @@ import requests
 from roboflow.adapters.rfapi import RoboflowError
 from roboflow.config import API_URL
 
+DEFAULT_TIMEOUT = (10, 60)
+
 
 class DeviceApiError(RoboflowError):
     """Raised when a device API call returns a non-success status."""
@@ -96,7 +98,7 @@ def _raise_for_status(response: requests.Response) -> None:
 
 def list_devices(api_key: str, workspace: str) -> Dict[str, Any]:
     """``GET /:workspace/devices/v2`` — returns the parsed JSON response."""
-    response = requests.get(_build_url(workspace, "", api_key))
+    response = requests.get(_build_url(workspace, "", api_key), timeout=DEFAULT_TIMEOUT)
     _raise_for_status(response)
     return response.json()
 
@@ -125,14 +127,18 @@ def create_device(
     if source_device_id is not None:
         # Body field is camelCase per docs/api/deployments/overview.md
         body["sourceDeviceId"] = source_device_id
-    response = requests.post(_build_url(workspace, "", api_key), json=body)
+    response = requests.post(
+        _build_url(workspace, "", api_key),
+        json=body,
+        timeout=DEFAULT_TIMEOUT,
+    )
     _raise_for_status(response)
     return response.json()
 
 
 def get_device(api_key: str, workspace: str, device_id: str) -> Dict[str, Any]:
     """``GET /:workspace/devices/v2/:deviceId``."""
-    response = requests.get(_build_url(workspace, f"/{device_id}", api_key))
+    response = requests.get(_build_url(workspace, f"/{device_id}", api_key), timeout=DEFAULT_TIMEOUT)
     _raise_for_status(response)
     return response.json()
 
@@ -144,7 +150,7 @@ def get_device_config(api_key: str, workspace: str, device_id: str) -> Dict[str,
         The response can include ``environment_variables`` and integration
         credentials. Treat the returned dict as sensitive.
     """
-    response = requests.get(_build_url(workspace, f"/{device_id}/config", api_key))
+    response = requests.get(_build_url(workspace, f"/{device_id}/config", api_key), timeout=DEFAULT_TIMEOUT)
     _raise_for_status(response)
     return response.json()
 
@@ -164,7 +170,8 @@ def get_device_config_history(
             f"/{device_id}/config/history",
             api_key,
             query={"limit": limit, "cursor": cursor},
-        )
+        ),
+        timeout=DEFAULT_TIMEOUT,
     )
     _raise_for_status(response)
     return response.json()
@@ -172,14 +179,17 @@ def get_device_config_history(
 
 def list_device_streams(api_key: str, workspace: str, device_id: str) -> Dict[str, Any]:
     """``GET /:workspace/devices/v2/:deviceId/streams``."""
-    response = requests.get(_build_url(workspace, f"/{device_id}/streams", api_key))
+    response = requests.get(_build_url(workspace, f"/{device_id}/streams", api_key), timeout=DEFAULT_TIMEOUT)
     _raise_for_status(response)
     return response.json()
 
 
 def get_device_stream(api_key: str, workspace: str, device_id: str, stream_id: str) -> Dict[str, Any]:
     """``GET /:workspace/devices/v2/:deviceId/streams/:streamId``."""
-    response = requests.get(_build_url(workspace, f"/{device_id}/streams/{stream_id}", api_key))
+    response = requests.get(
+        _build_url(workspace, f"/{device_id}/streams/{stream_id}", api_key),
+        timeout=DEFAULT_TIMEOUT,
+    )
     _raise_for_status(response)
     return response.json()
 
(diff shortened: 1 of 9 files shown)


NVergunst-ROBO (commented):
leaving a couple notes

roboflow/adapters/devicesapi.py
```diff
@@ -0,0 +1,272 @@
...
+            raise DeviceAuthError(message or "Forbidden", status_code=code)
+        raise DeviceNotFoundError(message or "Not found", status_code=code)
+    if code == 429:
+        raise DeviceRateLimitedError(message or "Rate limited", status_code=code)
+    raise DeviceApiError(message or f"HTTP {code}", status_code=code)
+
+
+def list_devices(api_key: str, workspace: str) -> Dict[str, Any]:
+    """``GET /:workspace/devices/v2`` — returns the parsed JSON response."""
+    response = requests.get(_build_url(workspace, "", api_key))
```
NVergunst-ROBO: These requests calls do not pass timeout=, so a slow or stuck API can hang the CLI and any SDK caller forever. A default connect/read timeout would make behaviour better

roboflow/adapters/devicesapi.py
```diff
@@ -0,0 +1,272 @@
...
+        if isinstance(err, dict):
+            message = err.get("message") or response.text
+            raw_type = err.get("type")
+            error_type = raw_type if isinstance(raw_type, str) else None
+        elif isinstance(err, str):
+            message = err
+        else:
+            message = response.text
+    except Exception:  # noqa: BLE001
+        message = response.text
```
NVergunst-ROBO, line 93: response.text can be huge, maybe add a truncate for max length?

roboflow/cli/handlers/device.py
```diff
@@ -0,0 +1,548 @@
...
+
+    resolved = _resolve_ws_and_key(args)
+    if not resolved:
+        return
+    ws, api_key = resolved
+
+    try:
+        result = devicesapi.list_devices(api_key, ws)
+    except Exception as exc:  # noqa: BLE001
+        output_error(args, str(exc), hint=_hint_for(exc), exit_code=_exit_code_for(exc))
```
NVergunst-ROBO, line 249: str(exc) on some requests failures can include the request URL, and the key is in the query string. Perhaps a secrets redaction would avoid accidentally printing the key in the terminal.

roboflow/cli/handlers/device.py
```diff
@@ -0,0 +1,548 @@
...
+        return
+    ws, api_key = resolved
+
+    try:
+        config = devicesapi.get_device_config(api_key, ws, args.device_id)
+    except Exception as exc:  # noqa: BLE001
+        output_error(args, str(exc), hint=_hint_for(exc), exit_code=_exit_code_for(exc))
+        return
+
+    output(args, config)
```
NVergunst-ROBO, line 369: Same as above, config can include env vars and integration secrets, maybe redact

NVergunst-ROBO (dismissed):
A couple nits

roboflow/adapters/devicesapi.py
rvirani1: Addressed in 7ce03eb by adding a shared DEFAULT_TIMEOUT=(10, 60) to every devicesapi requests.get/post call, plus a unit test that exercises all adapter request paths and asserts the timeout is passed.

roboflow/adapters/devicesapi.py
rvirani1: Addressed in aaa1ac2: added a 1024-char `_truncate` cap in `_raise_for_status` so the message that lands in `DeviceApiError(...)` (and therefore `str(exc)`) is bounded. Anything past the cap gets `…[truncated]`. Tested with a 10KB synthetic body in `tests/test_device.py::test_500_truncates_huge_response_body`.

roboflow/cli/handlers/device.py
rvirani1: Good catch — addressed in aaa1ac2.

Two things going on:
1. `_sanitize_credentials` in `roboflow/cli/_output.py` already runs on every error message routed through `output_error`, but the old regex `api_key=[A-Za-z0-9_]+` missed keys with dashes (e.g. UUID-shaped or hyphenated keys). Tightened to `api_key=[^\s&"'\\<>]+`.
2. The handlers all funnel exceptions through `output_error(args, str(exc), ...)`, so the sanitization now applies uniformly to anything `requests` puts in the URL.

Regression test: `tests/cli/test_device_handler.py::TestDeviceErrorRedaction::test_connection_error_strips_api_key_from_url` — feeds a `RuntimeError("...?api_key=secret-key-with-DASH_und_123")` through `_get`, asserts the secret never appears on stderr and the placeholder `api_key=***` does.

roboflow/cli/handlers/device.py
rvirani1: Pushed back on outright redaction in aaa1ac2 — here's the reasoning, and what I did instead.

**Why we shouldn't redact in the SDK:** `GET /:workspace/devices/v2/:deviceId/config` is documented in `docs/api/deployments/overview.md` as a passthrough of the full Firestore config doc, with the explicit note "treat the response as sensitive." If the SDK silently strips fields, we break round-trip use cases — backup/restore, config diffing across devices, migration tooling all expect what they see to match what the API returned. We'd also drift from every other client (UI, future light SDK, raw curl) that doesn't redact. If we want redaction, it belongs server-side in the mappers (which is where `sanitizeSource` already redacts stream credentials uniformly for everyone).

**What I did instead:** `roboflow device config <id>` now prints a `WARNING: Device config may contain environment variables, API keys, and integration credentials. Do not paste this output into chats, tickets, screenshots, or shared logs.` to **stderr** in interactive (text, non-quiet) mode. JSON mode (machine consumers) and `--quiet` are intentionally silent so we don't break piping. The data payload is byte-faithful to the API in both modes.

Regression tests in `tests/cli/test_device_handler.py::TestDeviceConfigWarning`:
- `test_text_mode_prints_warning_to_stderr` — confirms warning appears
- `test_json_mode_no_warning_and_no_payload_mutation` — confirms JSON mode is silent **and** the dumped payload equals the API response exactly (no SDK-side redaction)

Happy to escalate to a server-side redaction discussion if you'd rather see fields like `environment_variables` masked uniformly across clients — that's a roboflow/roboflow change in `services/device/externalApi/mappers.js`.

NVergunst-ROBO (approved):
👍👍

NVergunst-ROBO:
Sweet 👍

PR #478, opened 2026-05-12
Include keypoints_metadata.json when uploading model

grzegorz-roboflow:
## What does this PR do?

Include keypoints_metadata.json when uploading model, so users can upload keypoint models 

## Type of Change

- Bug fix (non-breaking change that fixes an issue)

## Testing

Tested locally

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [x] I have updated the documentation accordingly (if applicable)

## Additional Context

N/A

### DIFF
--- a/roboflow/__init__.py
+++ b/roboflow/__init__.py
@@ -21,7 +21,7 @@ except ImportError:
     CLIPModel = None  # type: ignore[assignment,misc]
     GazeModel = None  # type: ignore[assignment,misc]
 
-__version__ = "1.3.9"
+__version__ = "1.3.10"
 
 
 def check_key(api_key, model, notebook, num_retries=0):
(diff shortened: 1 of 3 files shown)


tonylampada (dismissed):
LGTM

there's improvement suggestion from [🤖 review](https://gist.github.com/tonylampada/d86f86d713fea9864d8cdafbfdb7426d) that might be worth looking into:

> a user deploying a pose model from a directory that lacks keypoints_metadata.json gets a successful upload acknowledgment but a broken deployed model — with no feedback about what went wrong.

> This is a "should fix before merge" concern, not a blocker, because the PR already improves the status quo

PR #487, opened 2026-05-29
fix(rfdetr): accept dict-form checkpoint args in classname extraction

leeclemnet:
## What does this PR do?

In `get_classnames_txt_for_rfdetr`, normalizes `checkpoint["args"]` to a dict instead of calling `vars()` on it directly, and adds tests for both dict and `Namespace` shapes.

`vars(checkpoint["args"])` raises `TypeError: vars() argument must have __dict__ attribute` when `args` is a plain dict rather than an `argparse.Namespace`. rf-detr EMA checkpoints (e.g. `checkpoint_best_ema.pth`) store `args` as a dict.

Today this is masked when a `class_names.txt` already exists in the upload dir (the function short-circuits before reaching `vars()`), but it breaks for any rf-detr checkpoint with dict-form args that relies on extracting class names from the checkpoint.

**Related Issue(s):** Companion to the server-side conversion fix roboflow/roboflow-model-conversion#93, which hit the same dict-vs-Namespace assumption on `args.resolution` and caused real prod upload failures (Medra workspace, `rfdetr-seg-medium`).

## Type of Change

- Bug fix (non-breaking change that fixes an issue)

## Testing

- [x] I have tested this change locally
- [x] I have added/updated tests for this change

**Test details:**

`python -m unittest tests.util.test_model_processor` → 14 tests pass, including new `GetClassnamesTxtForRfdetrTest` covering both `args` shapes. The `test_dict_args` case reproduces the `vars()` TypeError against the old code. `ruff format --check` and `ruff check` clean on the changed files; pre-commit hooks passed on push.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [ ] I have updated the documentation accordingly (if applicable)

## Additional Context

The two normalizations point opposite directions on purpose: this site needs a **dict** (downstream does `args["class_names"]`), while the conversion site needs a **Namespace** (downstream does `args.resolution`). Each normalizes toward what its own caller expects.

### DIFF
--- a/requirements-slim.txt
+++ b/requirements-slim.txt
@@ -6,10 +6,7 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 requests_toolbelt
 filetype
-# typer 0.26 vendors click as typer._click and drops the external click dep, breaking
-# roboflow.cli imports (test_slim_compat imports it). 0.25.x still uses external click.
-typer>=0.12.0,<0.26
-# CLI imports click directly; declare it explicitly rather than relying on typer's transitive dep.
+typer>=0.12.0,<0.26  # 0.26 vendors click, dropping the external dep the CLI imports
 click>=8.0
 python-dateutil
 python-dotenv
--- a/requirements-slim.txt
+++ b/requirements-slim.txt
@@ -6,7 +6,11 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 requests_toolbelt
 filetype
-typer>=0.12.0
+# typer 0.26 vendors click as typer._click and drops the external click dep, breaking
+# roboflow.cli imports (test_slim_compat imports it). 0.25.x still uses external click.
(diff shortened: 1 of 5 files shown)


requirements-slim.txt
```diff
@@ -6,7 +6,8 @@ tqdm>=4.41.0
 PyYAML>=5.3.1
 requests_toolbelt
 filetype
-typer>=0.12.0
+typer>=0.12.0,<0.26  # 0.26 vendors click, dropping the external dep the CLI imports
+click>=8.0
```
leeclemnet, line 10: CI broken without this

probicheaux (dismissed):

probicheaux (approved):

PR #502, opened 2026-06-30
fix: return single_upload result from Project.upload()

### LINKED ISSUE #254: Project#upload does not return anything
Why does `Project#single_upload` return a response while `upload` returns `nil` ?

Is this an oversight or expected behaviour?

davidnichols-ops:
## Summary
- `Project.upload()` was discarding the return value of `single_upload()`, always returning `None` even on successful uploads. This made it impossible for callers to inspect the upload response (image id, timing, retry counts) without calling `single_upload()` directly.
- Now returns the `single_upload()` result dict for single-file uploads (keys: `image`, `annotation`, `upload_time`, `annotation_time`, `upload_retry_attempts`, `annotation_upload_retry_attempts`).
- For directory uploads, returns a list of such dicts — one per successfully uploaded image. Skipped (non-image) files are excluded from the list.
- Existing callers that ignore the return value are completely unaffected — this is a backwards-compatible addition.

Closes #254.

#### Test plan
- [x] `python -m unittest` passes — 731 tests, OK (skipped=1)
- [x] New test `test_upload_single_file_returns_result` verifies single-file upload returns a dict with the expected image id
- [x] New test `test_upload_directory_returns_list_of_results` verifies directory upload returns a list of dicts with correct count
- [x] Updated existing `test_project_methods` in `test_queries.py` to assert the new return value (was asserting `None`)
- [x] `ruff check` and `ruff format --check` pass on all modified files

### DIFF
--- a/roboflow/core/project.py
+++ b/roboflow/core/project.py
@@ -409,6 +409,12 @@ class Project:
             metadata (dict, optional): custom key-value metadata to attach to the image.
                 Example: {"camera_id": "cam001", "location": "warehouse"}
 
+        Returns:
+            For a single image: the dict returned by ``single_upload`` (keys: ``image``,
+            ``annotation``, ``upload_time``, ``annotation_time``, ``upload_retry_attempts``,
+            ``annotation_upload_retry_attempts``). For a directory: a list of such dicts,
+            one per successfully uploaded image. Skipped (non-image) files are excluded.
+
         Example:
             >>> import roboflow
 
@@ -445,7 +451,7 @@ class Project:
                     )
                 )
 
-            self.single_upload(
+            return self.single_upload(
                 image_path=image_path,
                 annotation_path=annotation_path,
                 hosted_image=hosted_image,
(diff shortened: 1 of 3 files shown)


davidnichols-ops:
Hi @digaobarbosa @leeclemnet — gentle follow-up on this PR. It fixes a bug where `Project.upload()` discards the `single_upload` result and returns `None` instead of the upload response. Small change (55 lines), mergeable, no conflicts. Ready for review when you have bandwidth.

roboflow/core/project.py
```diff
@@ -445,7 +451,7 @@ def upload(
                     )
                 )
 
-            self.single_upload(
+            return self.single_upload(
```
digaobarbosa: I think we should return a list in both single and folder upload cases. Just to keep the return type consistency.

digaobarbosa (changes requested):
Thanks for the contribution @davidnichols-ops , can you address the comment I added?

davidnichols-ops:
Thanks for the review @digaobarbosa! Good call on the return type consistency.

Updated in 5698c09: `upload()` now returns `list[dict]` in both cases — single-file uploads return `[single_upload_result]` instead of a bare dict. The docstring and tests have been updated to match.

All 29 tests in `test_project.py` and 8 in `test_queries.py` pass, lint is clean.

digaobarbosa (approved):
LGTM

Issue #498, opened 2026-06-30
CI typecheck fails: numpy 2.5.0 stubs use PEP 695 `type` statements, incompatible with `mypy python_version = "3.10"`


davidnichols-ops:
# CI typecheck fails: numpy 2.5.0 stubs use PEP 695 `type` statements, incompatible with `mypy python_version = "3.10"`

## Summary

`make check_code_quality` (the `mypy roboflow` step in CI) started failing on fresh installs after numpy 2.5.0 was released on 2026-06-21. numpy 2.5.0 dropped Python 3.11 support specifically so its bundled type stubs could move to PEP 695 `type` statement syntax (Python 3.12+). This repo's mypy config targets Python 3.10, so mypy cannot parse the new stubs and errors out — regardless of the mypy version or the Python interpreter running CI.

This is a numpy × mypy version interaction, not a bug in roboflow code. The repo has no upper bound on numpy in its dev/test extra, so `pip install ".[dev]"` now resolves numpy 2.5.0 and breaks the typecheck gate on every push and PR.

## Reproduce

```bash
python3 -m venv env && source env/bin/activate
pip install -e ".[dev]"        # pulls numpy 2.5.0
make check_code_quality        # ruff passes, mypy fails
```

CI matrix affected: all of `ubuntu-latest` / `windows-latest` × `3.10` / `3.11` / `3.12` / `3.13`, since the mypy *target* (not the interpreter) is pinned to 3.10.

## Root cause

- `pyproject.toml`:
  ```toml
  [tool.mypy]
  python_version = "3.10"
  ```
- `requirements.txt`:
  ```
  numpy>=1.18.5
  ```
- `setup.py` lists `numpy` in the dev extra with no upper bound.

numpy 2.5.0 release notes confirm the move to PEP 695 stubs (and that dropping Python 3.11 was the enabling step): https://numpy.org/devdocs/release/2.5.0-notes.html

An independent project hit the same break on 2026-06-22 and fixed it by capping `numpy<2.5` in their dev extra: https://github.com/bertravacca/ev-flow/commit/3abab0d956d0aa7aaeaa5b2dfbdd8f4382906789 — their note confirms mypy 2.0 and 2.1 both fail under a py3.10 target, so this is not a mypy-version issue.

## Suggested fix

Cap `numpy<2.5` in the **dev/test extra only** (`setup.py`), leaving the runtime `requirements.txt` constraint (`numpy>=1.18.5`) untouched so end users are not affected:

```python
[...]

PR #495, opened 2026-06-20
feat(api-keys): roboflow api-key CLI

yeldarby:
Adds a `roboflow api-key` CLI command group (list / get / create / update / revoke / protect / disable / publishable) wrapping the new REST endpoints.

- Secret shown once on create; non-secret `keyId` handles; `--scope` / `--metadata` on create + update; protected-key 409 and plan-feature 403 hints; consistent exit codes and `--json` envelopes.
- Tests in `tests/cli/test_api_key.py`; version bump.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #65, opened 2022-10-10
Switch github action target branch from master to main

broox:
# Description

Github actions were not being run after renaming the base branch from `master` to `main`. This should fix that.

## Type of change

Please delete options that are not relevant.

-   [x] Bug fix (non-breaking change which fixes an issue)

## How has this change been tested, please provide a testcase or example of how you tested the change?

I opened this PR and saw that the actions are working again.

PR #503, opened 2026-07-02
refactor: expose a public custom-weights packaging helper

lrosemberg:
## What this PR does

Makes the SDK the single owner of custom-weights packaging. It exposes a public, non-interactive helper that both the SDK deploy flows and external callers (such as the Roboflow MCP server) use, so the packaging logic lives in one place instead of being copied per caller.

The old packaging entry point could not be reused outside the CLI: `process()` prompts with `input()`, can call `sys.exit(1)`, prints to stdout, and writes artifacts into the caller's model directory. This PR extracts a clean, side-effect-free helper and keeps the CLI behavior behind a thin compatibility layer.

```
Before:                          After:
Version.deploy(...)              package_custom_weights(...)   (only packages, no I/O side effects)
Workspace.deploy_model(...)        ├─ Version.deploy(...)            package + SDK upload
  packaging logic inlined          ├─ Workspace.deploy_model(...)    package + SDK upload
  in each deploy flow              └─ external callers               package + their own upload
```

## New public API: `roboflow/util/model_processor.py`

`package_custom_weights(model_type, model_path, filename="weights/best.pt", *, build_dir=None, allow_dependency_mismatch=False, allow_size_mismatch=False) -> ModelUploadBundle`

- **Non-interactive and side-effect free on `model_path`.** It never prompts, prints, exits, or writes into the source directory. Heavy deps (`torch`, `ultralytics`) are imported lazily, only for the families that need them.
- **Build directory ownership.** `build_dir=None` packages into a fresh `mkdtemp()` owned by the returned bundle (`bundle.cleanup()` removes it). Passing an explicit `build_dir` uses it as-is and leaves ownership to the caller. The old "write into `model_path`" behavior lives only behind the compatibility `process()`.
- **Returns** `ModelUploadBundle` with `archive_path`, `build_dir`, `owns_build_dir`, the resolved `model_type`, and collected `warnings`.

**Typed error contract:** `ModelPackagingError` (base), with `UnsupportedModelError`, `TaskMismatchError`, `MissingFileError`, `MissingDependencyError`, `DependencyMismatchError`, and `SizeMismatchError`. Every `ModelPackagingError` is a user-correctable input problem; anything else escaping the helper is a bug. Each subclass also inherits the builtin type the old code raised (`ValueError` / `FileNotFoundError` / `RuntimeError`), so existing `except` blocks keep working.

**Validations** (some new, some already in this module):
- **YOLO size-suffix inference and checking.** Roboflow rejects bare family names (`yolov8`, `yolov11`, ...). The size is inferred from the model yaml's `scale` letter or the `depth_multiple`/`width_multiple` pair; a missing suffix is filled in (`yolov8` becomes `yolov8n`, with a warning), and a declared size that conflicts with the checkpoint raises `SizeMismatchError` naming the size that fits, instead of failing server-side later.
- **RF-DETR variant checking.** The checkpoint's position-encoding grid (from `positional_encoding_size`, `resolution / patch_size`, or the backbone `position_embeddings` tensor) is reconciled against the requested variant; a known mislabel raises pointing at the matching variant, an unknown or custom grid warns and proceeds.
- Cleaner errors for missing files, malformed checkpoints, and missing `torch`/`ultralytics`.

**Prefix-based family gate:** `model_type.startswith(SUPPORTED_MODELS)` rather than substring containment, so an invalid type like `foo-yolov8n` is rejected up front instead of being half-recognized and rejected by the backend after upload.

## Backwards compatibility

- `process(model_type, model_path, filename)` is kept as a thin wrapper that delegates to `package_custom_weights_interactive(build_dir=model_path)`: it packages into `model_path` (same intermediate artifacts and archive landing there as before), prints warnings, keeps the historical print-and-confirm on dependency/size mismatches, and returns `(zip_file_name, model_type)`.
- `Version.deploy` and `Workspace.deploy_model` keep their signatures. They call `package_custom_weights_interactive`, which preserves the historical CLI UX: warnings are printed, and dependency/size mismatches ask for confirmation before retrying with the matching override.
- `validate_model_type_for_project`, `task_of_model_type`, `_detect_yolo_task`, `_detect_rfdetr_task`, `get_classnames_txt_for_rfdetr`, and `maybe_prepend_dummy_class` keep their signatures. `yolo26-sem` and semantic-segmentation handling are unchanged.

## Hardening

Edge cases found while testing against real checkpoints, each fixed with a test:

- **RF-DETR re-deploy `SameFileError`.** In the legacy `Version.deploy` / `Workspace.deploy_model` flow `build_dir == model_path`, so a top-level `weights.pt` was `shutil.copy`'d onto itself and raised `SameFileError` (an `OSError`, outside the packaging error contract). The self-copy is now skipped when source and destination resolve to the same path.
- **Bare-family infinite retry loop.** When a bare family name (e.g. `yolov10`) had no inferable size, `_resolve_yolo_size` raised `SizeMismatchError` unconditionally, even under `allow_mismatch`; the interactive retry then re-prompted forever. It now converges (packages the bare name with a warning once `allow_size_mismatch` is set), matching the old deploy behavior.
- **Silent wrong checkpoint on a typo'd filename.** An explicit, non-default `filename` that doesn't exist no longer falls back to discovering some other `.pt`/`.pth` in the directory (which would upload the wrong weights); it raises `MissingFileError`. The default path still discovers, as before.
- **Corrupt checkpoint `args`.** A checkpoint storing `args` as a bare scalar/list raised a raw `TypeError` from `vars()`; a small `_checkpoint_args_as_dict` helper coerces those to `{}` so the error contract holds.

**Checkpoint-completeness guards.** Roboflow's server-side conversion reads metadata that a bare inference state_dict (e.g. `{"model": <tensors>}` with nothing else) does not carry, so such a file used to package and upload fine and then fail conversion with an opaque error like `KeyError: 'args'`. Packaging now rejects these up front with an actionable message:
- **RF-DETR:** the checkpoint must contain `args` (class names, class count, model config); a bare state_dict is rejected instead of failing server-side.
- **YOLO-NAS:** the checkpoint must contain `processing_params.class_names`; a bare state_dict is rejected.
- **YOLO args coercion:** `_filtered_args` now routes through `_checkpoint_args_as_dict`, so a corrupt `model.args` coerces to `{}` rather than raising a raw `TypeError`.

Deferred (pre-existing behavior, not introduced here): the `_process_yolo` branches dereference `checkpoint["train_args"]`, `model_instance.nc`, and `model_instance.yaml["nc"]` directly, which can raise raw exceptions on non-standard or reconstructed Ultralytics checkpoints. These only bite atypical inputs and guarding them touches the happy path across Ultralytics versions, so they are left for a separate hardening pass.

## Intentional behavior changes (all previously interactive dead ends)

- Declining a dependency-version prompt in the deploy flows now re-raises a typed error instead of `sys.exit(1)`.
- Hugging Face uploads with missing tokenizer/preprocessor sidecar files now fail with a clear `MissingFileError` instead of prompting y/n and calling `exit(1)`.
- RF-DETR checkpoint discovery: the default `filename` still falls back to the first top-level `.pt`/`.pth` (sorted for determinism) with a warning, preserving historical behavior; an explicit non-default `filename` that is missing is now a hard error rather than a silent fallback (see hardening above).

## Tests

- Packaging tests in `tests/util/test_model_processor.py` cover size/variant resolution, the error contract, and RF-DETR class names.
- Contract tests: `package_custom_weights` fails the test if `input()` or `sys.exit()` is ever called; the source directory is byte-identical before and after packaging with a temp build dir; the owned temp dir is removed on failure; `process()` still writes the archive into `model_path`, keeps the interactive confirm, and returns the old tuple.
- Regression tests for each hardening fix above.
- `python -m unittest` (832 tests), `ruff format --check`, `ruff check`, and `mypy` all green on the CI matrix (Ubuntu + Windows, Python 3.10 to 3.13).

## Follow-ups and coordination

- roboflow/roboflow-mcp#65 switches to calling `package_custom_weights` once this lands in a PyPI release. For the MCP to install the SDK, the exact `opencv-python-headless==4.10.0.84` pin in `requirements.txt` also needs to relax; #501 (still open) does that. Until both land, the MCP PR pins this branch by git ref.
- #496, #493, #488, and #478 all touch this module and will need small rebases. The constants (`SUPPORTED_RFDETR_TYPES`, ...) and the per-family `_process_*` function names were kept to make those rebases mechanical.

### DIFF
--- a/roboflow/util/model_processor.py
+++ b/roboflow/util/model_processor.py
@@ -102,6 +102,8 @@ SUPPORTED_RFDETR_TYPES = (
     "rfdetr-seg-2xlarge",
 )
 
+DEFAULT_WEIGHTS_FILENAME = "weights/best.pt"
+
 # YOLO families Roboflow rejects without a size suffix (e.g. `yolov8` must be
 # `yolov8n`/`yolov8s`/...). Legacy yolov5/7/9 go through the opt.yaml path and are
 # intentionally excluded.
@@ -237,6 +239,20 @@ def task_of_model_type(model_type: str) -> str:
     return TASK_DET
 
 
+def _checkpoint_args_as_dict(raw_args: Any) -> dict[str, Any]:
+    """Coerce a checkpoint's ``args`` (dict, argparse.Namespace, or None) to a dict.
+
+    A corrupt checkpoint may store ``args`` as a scalar or list; those have no
+    meaningful attributes, so treat them as no-args rather than letting
+    ``vars()`` raise a bare ``TypeError`` outside the ModelPackagingError contract.
+    """
+    if isinstance(raw_args, dict):
+        return raw_args
+    if hasattr(raw_args, "__dict__"):
+        return dict(vars(raw_args))
+    return {}
+
+
 def validate_model_type_for_project(model_type: str, project_type: str, project_id: str) -> None:
     """Raise TaskMismatchError if model_type's task doesn't match the Roboflow project type."""
     expected = {
@@ -259,7 +275,7 @@ def validate_model_type_for_project(model_type: str, project_type: str, project_
 def package_custom_weights(
     model_type: str,
     model_path: str,
-    filename: str = "weights/best.pt",
+    filename: str = DEFAULT_WEIGHTS_FILENAME,
     *,
     build_dir: str | Path | None = None,
     allow_dependency_mismatch: bool = False,
@@ -331,7 +347,7 @@ def package_custom_weights(
 def package_custom_weights_interactive(
     model_type: str,
     model_path: str,
-    filename: str = "weights/best.pt",
+    filename: str = DEFAULT_WEIGHTS_FILENAME,
     *,
     build_dir: str | Path | None = None,
 ) -> ModelUploadBundle:
@@ -547,6 +563,13 @@ def _resolve_yolo_size(
 
     if inferred is None:
         if not provided:
+            if allow_mismatch:
+                warnings.append(
+                    f"Could not infer a size for '{model_type}' from the checkpoint; "
+                    f"uploading the bare family name as requested. Roboflow may reject it "
+                    f"if it requires an explicit size."
+                )
+                return model_type
             raise SizeMismatchError(
                 f"model_type '{model_type}' is missing a size suffix and the size "
                 f"could not be inferred from the checkpoint. Specify it explicitly, "
@@ -731,7 +754,7 @@ def _detect_rfdetr_task(checkpoint: Any) -> str | None:
     raw_args = checkpoint.get("args")
     if raw_args is None:
         return None
-    args = raw_args if isinstance(raw_args, dict) else vars(raw_args)
+    args = _checkpoint_args_as_dict(raw_args)
     segmentation_head = args.get("segmentation_head")
     if segmentation_head is True:
         return TASK_SEG
@@ -749,13 +772,7 @@ def _rfdetr_checkpoint_pe_size(checkpoint: Any) -> int | None:
     """
     if not isinstance(checkpoint, dict):
         return None
-    raw_args = checkpoint.get("args")
-    if isinstance(raw_args, dict):
-        args = raw_args
-    elif raw_args is not None:
-        args = dict(vars(raw_args))
-    else:
-        args = {}
+    args = _checkpoint_args_as_dict(checkpoint.get("args"))
 
     pe = args.get("positional_encoding_size")
     if isinstance(pe, int) and pe > 0:
@@ -847,15 +864,23 @@ def _resolve_rfdetr_variant(
 
 
 def _find_rfdetr_checkpoint(model_path: Path, filename: str, warnings: list[str]) -> Path:
-    """Locate the rf-detr checkpoint, preserving the historical discovery behavior.
+    """Locate the rf-detr checkpoint.
 
-    The requested ``filename`` wins when it exists. Otherwise fall back to the
-    first top-level .pt/.pth file (sorted for determinism), which is how rf-detr
-    uploads located the checkpoint before ``filename`` was honored for them.
+    An explicitly-requested ``filename`` (anything other than the default) must
+    exist: falling back to a different checkpoint on a typo would silently
+    package the wrong weights. Only the default path falls back to discovering
+    the first top-level .pt/.pth file (sorted for determinism), preserving how
+    rf-detr uploads located the checkpoint before ``filename`` was honored.
     """
     requested_file = model_path / filename
     if requested_file.exists():
         return requested_file
+
+    if filename != DEFAULT_WEIGHTS_FILENAME:
+        raise MissingFileError(
+            f"RF-DETR weights file '{requested_file}' was not found. Set filename to the "
+            f"checkpoint's exact .pt or .pth path relative to model_path."
+        )
 
     discovered = sorted(path for path in model_path.iterdir() if path.is_file() and path.suffix in {".pt", ".pth"})
     if not discovered:
@@ -874,13 +899,7 @@ def _write_rfdetr_class_names(model_path: Path, build_dir: Path, checkpoint: Any
         class_names = class_names_path.read_text().splitlines()
     else:
         raw_args = checkpoint.get("args") if isinstance(checkpoint, dict) else None
-        if raw_args is None:
-            args: dict[str, Any] = {}
-        elif isinstance(raw_args, dict):
-            args = raw_args
-        else:
-            args = dict(vars(raw_args))
-        class_names = args.get("class_names") or []
+        class_names = _checkpoint_args_as_dict(raw_args).get("class_names") or []
         if not class_names:
             raise MissingFileError(
                 f"No class_names.txt file found in '{model_path}', and the RF-DETR "
@@ -924,7 +943,11 @@ def _process_rfdetr(
 
     model_type = _resolve_rfdetr_variant(model_type, checkpoint, warnings, allow_size_mismatch)
 
-    shutil.copy(checkpoint_path, build_dir / "weights.pt")
+    weights_dest = build_dir / "weights.pt"
+    # In the legacy deploy flow build_dir is model_path, so a checkpoint already
+    # named weights.pt is its own destination; copying would raise SameFileError.
+    if checkpoint_path.resolve() != weights_dest.resolve():
+        shutil.copy(checkpoint_path, weights_dest)
     _write_rfdetr_class_names(model_path, build_dir, checkpoint)
     archive_path = build_dir / "roboflow_deploy.zip"
     _write_zip(
--- a/roboflow/util/model_processor.py
+++ b/roboflow/util/model_processor.py
@@ -285,6 +285,11 @@ def _resolve_within_source(source_dir: Path, filename: str) -> Path:
             f"filename '{filename}' resolves outside model_path '{source_dir}'. "
             "It must point to a checkpoint inside the model directory."
         )
+    # '' / '.' resolve to model_path itself and a subdirectory passes the
+    # containment check; all would reach torch.load(<dir>) and leak a raw
+    # IsADirectoryError outside the ModelPackagingError contract.
+    if resolved.is_dir():
+        raise ModelPackagingError(f"filename '{filename}' must point to a checkpoint file, not a directory.")
     return resolved
 
 
--- a/roboflow/util/model_processor.py
+++ b/roboflow/util/model_processor.py
@@ -84,23 +84,32 @@ SUPPORTED_HUGGINGFACE_TYPES = (
     "paligemma2-3b-pt-896-peft",
 )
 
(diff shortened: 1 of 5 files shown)


lrosemberg:
CI note: the `build (ubuntu-latest, 3.13)` failure was the pre-existing numpy 2.5 / mypy stub issue (#498), not this change; the matrix fail-fast then canceled the other jobs (3.11 had already passed the quality gate). I cherry-picked the same dev-extra `numpy<2.5` cap that #495 carries in commit a4c0891 so this PR's CI can run green in the meantime. If #495 merges first, this line rebases away cleanly.

lrosemberg:
Follow-up hardening from a review pass on the packaging helper (all with tests):

- **rf-detr re-deploy `SameFileError`**: in the legacy `Version.deploy` / `Workspace.deploy_model` flow `build_dir == model_path`, so a top-level `weights.pt` was copied onto itself. The self-copy is now skipped.
- **Bare YOLO family infinite loop**: `_resolve_yolo_size` raised unconditionally when a bare family name (e.g. `yolov10`) had no inferable size, even under `allow_mismatch`; the interactive retry then looped forever. It now converges (packages the bare name with a warning) once `allow_size_mismatch` is set, matching the old deploy behavior.
- **Silent wrong checkpoint**: an explicit non-default `filename` that doesn't exist no longer falls back to discovering some other `.pt` in the directory (a typo would upload the wrong weights); it raises `MissingFileError`. The default path still discovers, as before.
- **Corrupt `args`**: `vars()` on a scalar/list `args` raised a bare `TypeError` outside the `ModelPackagingError` contract; a small `_checkpoint_args_as_dict` coerces those to `{}`.

Also rebased onto latest `main` and dropped the redundant `numpy<2.5` dev-extra commit (main already caps it in `requirements.txt` via #495) and the auto-format commit, so the history is single-author for the CLA check. `python -m unittest` green (832 tests).

lrosemberg:
Ready for review. This pulls the custom-weights packaging into a public, non-interactive `package_custom_weights` helper so the logic lives in one place, with a typed error contract, and keeps the old CLI behavior behind a thin compatibility wrapper (`process()`, `Version.deploy`, and `Workspace.deploy_model` are unchanged for callers).

Since the first version I hardened a handful of edge cases that showed up while testing against real checkpoints: a `SameFileError` on rf-detr re-deploy, an infinite retry loop on bare YOLO family names, a silent wrong-checkpoint on a typo'd filename, and a few bare-state-dict cases that used to fail server-side with an opaque `KeyError` (now caught up front with an actionable message). Each has a test.

CI is green across the matrix (832 tests, ruff, mypy), and the history is single-author so the CLA check passes. roboflow/roboflow-mcp#65 depends on this shipping to PyPI. Would appreciate a review when you have a minute.

lrosemberg:
Thanks for the joint review. Both points are addressed on the MCP side (roboflow/roboflow-mcp#65):

1. **Mutable SDK branch dependency** — `pyproject.toml` now pins the SDK at the immutable commit `29d1aec3fcf7d8f75ebfb97275030f86e88ea135` instead of the `lean/package-custom-weights` branch, so the reviewed MCP behavior stays tied to the reviewed SDK behavior and can't drift on a rebase/delete. It will still be swapped for a released `roboflow>=x.y.z` before merge, per the pre-merge checklist.
2. **Versioned upload preflight** — for `version_number is not None`, the tool now calls `rf_api.versions_get(...)` before packaging, so a typo in `project_id`/`version_number` fails fast instead of after `torch.load` + zip on a large checkpoint. Post-packaging task validation stays, since it needs the SDK-resolved `bundle.model_type`. Added a regression test asserting a missing version never reaches `prepare_custom_weights_archive`.

`./dev/test.sh` is green (327 passed).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mike-Medvedev (changes requested):
Good extraction. `package_custom_weights()` is the right public surface for MCP and the deploy flows, and the interactive/compat split looks clean. CI is green.

Requesting changes on two contract issues:

1. **`filename` is not constrained under `model_path`.** Joins like `model_path / filename` follow absolute paths and `../`. MCP forwards caller `filename` into this helper, so on a hosted MCP host this is a path-read / wrong-weights risk, and it breaks the documented "relative to model_path" contract. After join, `resolve()` and require the checkpoint stays under the source dir; reject absolute `filename`. Add tests for `../` and absolute paths.

2. **YOLO happy path still leaks non-`ModelPackagingError` exceptions.** Accesses like `checkpoint["train_args"]`, `model_instance.yaml["nc"]`, and `model_instance.nc` can raise raw `KeyError` / `AttributeError`. MCP maps only `ModelPackagingError` to 400s, so atypical checkpoints become opaque 500s. RF-DETR / YOLO-NAS already got completeness guards; YOLO should match (at least missing `train_args` on the yolov10/11/12/26/`-cls` branch).

Also missing HF packaging tests for the prompt→`MissingFileError` behavior change. Minor follow-ups: resolve explicit `build_dir`, tighten RF-DETR multi-checkpoint discovery.

Once this ships to PyPI with `#501`, MCP `#65` can drop the git pin.

roboflow/util/model_processor.py
```diff
@@ -149,116 +489,211 @@ def _detect_yolo_task(model_instance) -> Optional[str]:
...
-            import torch
-            import ultralytics
+        ultralytics = _import_required_module("ultralytics", "pip install ultralytics")
 
-        except ImportError:
-            raise RuntimeError(
-                "The ultralytics python package is required to deploy yolo26"
-                " models. Please install it with `pip install ultralytics`"
-            )
+    checkpoint_path = model_path / filename
```
Mike-Medvedev, line 727: `model_path / filename` follows absolute paths and `../`. MCP forwards caller `filename` into this helper, so this can read outside the source dir (and absolute paths also bypass the "explicit missing filename" hard-error path).

After join, `resolve()` and require the checkpoint stays under the source dir; reject absolute `filename`. Worth a unit test for both cases.

roboflow/util/model_processor.py
```diff
@@ -267,78 +702,51 @@ def _process_yolo(model_type: str, model_path: str, filename: str) -> tuple[str,
             or model_type.startswith("yolo26")
         ):
             nc = model_instance.yaml["nc"]
-            args = model["train_args"]
+            args = checkpoint["train_args"]
```
Mike-Medvedev: Missing `train_args` / `.yaml` / `.nc` still raise raw `KeyError` / `AttributeError`. MCP maps only `ModelPackagingError` to 400s, so atypical checkpoints become opaque 500s.

RF-DETR / YOLO-NAS already got completeness guards; please match that here (at least missing `train_args` on this branch) so the public error contract holds.

Mike-Medvedev (commented):
Inline notes on the two contract blockers.

roboflow/util/model_processor.py
lrosemberg: Done. `package_custom_weights` now calls `_resolve_within_source(source_dir, filename)` before dispatch: it rejects an absolute `filename` and any join that `resolve()`s outside `model_path`, so the containment check now covers every processor (yolo / rfdetr / yolonas / hf) at a single choke point. This also closes the absolute-path bypass of the explicit-missing-filename hard error. Added unit tests for both the `../` escape and the absolute-path case.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

roboflow/util/model_processor.py
lrosemberg: Done. Added `_require_model_attr` / `_require_checkpoint_field` helpers so the YOLO metadata reads no longer leak raw `KeyError` / `AttributeError`. Missing `train_args` and a missing `nc` in `model.yaml` now raise `ModelPackagingError` on the yolov10/11/12/26/`-cls` branch, and `.yaml`, `.nc`, `.args` are guarded on the yolov8 and legacy yolov5/7/9 branches too, so the public 400 contract holds for atypical checkpoints. Added a focused test per field.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

roboflow/util/model_processor.py
lrosemberg: Follow-up: Windows CI caught that `Path.is_absolute()` is host-dependent (a POSIX `/etc/passwd` is not "absolute" on Windows, so it slipped past to the containment branch with a different message). Hardened the guard to test absoluteness under both conventions (`PurePosixPath` and `PureWindowsPath`), so a caller-supplied absolute path is rejected the same way regardless of which OS packaging runs on. Green on Linux and Windows now.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Mike-Medvedev (dismissed):
Looks good. The earlier blockers are addressed:

- `_resolve_within_source` rejects absolute/`../` filenames (cross-platform) and directory targets
- YOLO completeness goes through `_require_*` into `ModelPackagingError`
- HF packaging tests are in place

CI is green. LGTM.

digaobarbosa (approved):
LGTM

PR #518, opened 2026-08-24
Add Batch Processing CLI commands

stellasphere:
## Summary

Adds `roboflow batch` as a supported CLI surface for Batch Processing:

- Create a job from exact image IDs, a RoboQL query, or an explicit whole-library selection
- Read and search durable jobs
- Abort or restart a job with explicit confirmation
- Return stable JSON for scripts and automation
- Preserve the request ID through connection failures so retrying an uncertain request cannot create a second paid job
- Bound every HTTP request with explicit connect and read timeouts

The adapter uses the deployed canonical Platform contract. Asset Library creation uses `/asset-library/jobs`; list, status, abort, and restart use the shared `/jobs` lifecycle. Local folders remain an upload-first flow, so the CLI never implies the server can read a caller's filesystem.

Depends on the canonical operations shipped in [roboflow/roboflow#14693](https://github.com/roboflow/roboflow/pull/14693).

## Verification

Current head: `d97b26fef545b093791e3505d3ecdde1dd4669ad`.

- 37 focused adapter, CLI, and completion tests pass, plus 9 parameterized subtests
- Ruff format and lint pass
- Mypy passes
- `git diff --check` passes
- GitHub CI passes on Linux and Windows with Python 3.10 through 3.13, plus slim install, CodeQL, Socket, and pre-commit
- Rebased onto current `main` at `af1e7e2a0573a39160642aea6ae0d09b8d7d4b94`

### DIFF
--- a/CLI-COMMANDS.md
+++ b/CLI-COMMANDS.md
@@ -47,6 +47,27 @@ roboflow download my-workspace/my-project/3 -f coco   # alias
 roboflow infer photo.jpg -m my-project/3
 ```
 
+### Batch Process Asset Library images
+
+```bash
+# Exact reviewed selection (CPU is the product default):
+roboflow batch create --workflow inspect-defects --image-ids img_1,img_2
+
+# Or every current match for a structured RoboQL filter:
+roboflow batch create --workflow inspect-defects --query "tag:night-shift"
+
+# Monitor and control the durable job:
+roboflow batch status <job-id>
+roboflow batch list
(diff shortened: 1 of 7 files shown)


stellasphere:
/jarbas review

Please review exact head `c3275503877d93054bbde017c8bf61abba174c97`. The focused reliability regression suite and all GitHub checks pass.

Matvezy (dismissed):

Matvezy (approved):

